PR #142, opened 2023-01-24
Add missing tf.function parameters

### LINKED ISSUE #139: Missing tf.function parameters
https://github.com/ponder-lab/Hybridize-Functions-Refactoring/blob/cd6694ce0182f2e29d0ea182634a76dbcf238f1c/edu.cuny.hunter.hybridize.tests/test%20cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java#L733-L735

According to https://www.tensorflow.org/versions/r2.9/api_docs/python/tf/function, we are missing:
1. `experimental_relax_shapes`
2. `experimental_compile`

tatianacv:
Fixes #139.

## Commits

- [x] I slowed down and took a deep breadth before committing. Otherwise, I will start a pristine branch and reissue this pull request.
- [x] Before committing my changes, I checked my working directory carefully. I only staged the changes I wanted and ensured it was of the utmost quality.
- [x] I used `git diff` and `git status` to check my work carefully before committing.
- [x] I completely understand what I am committing and why.
- [x] I completely understand why things are the way they are.
- [x] The change I am proposing makes complete sense to me.
- [x] For the most part, the changes in this pull request correspond to the problem or task I am solving or performing. In other words, only the changes that are supposed to be there are the ones included in this pull request.

## Expectations

We should now have the params `experimental_compile` and `experimental_relax_shapes`. We added getters to those tf.function parameters and added an additional check for these in the tests.

## Testing

- [x] If this is a bug fix, I was able to reproduce the problem locally **before** I made any changes (can use `git stash` to temporarily reset the working directory). 
- [x] When I made (or put back) my changes, the problem I reproduced above was fixed.

Added another conditional statement to reflect the missing tf.function parameters.

## Final Checks

- [x] The changes I am proposing meet the expectations I have described above.
- [x] I have performed a self-review of my code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation if applicable.
- [x] I have added tests that prove my fix is effective or that my feature works if applicable.
- [x] New and existing unit tests pass locally with my changes.


khatchad (changes requested):

khatchad (changes requested):

khatchad (changes requested):

khatchad (changes requested):
Not sure how these convos got resolved.

khatchad (changes requested):

khatchad (dismissed):

khatchad:
Triage findings — recommending closure as superseded.

The intent of linked issue #139 ("we are missing detection of \`experimental_relax_shapes\` and \`experimental_compile\`") has been **substantially implemented** on the current branch. In `Function.java`:

- Constants `EXPERIMENTAL_COMPILE` (line 149) and `EXPERIMENTAL_RELAX_SHAPES` (line 155) are defined.
- The `computeParameterExistance` parser recognizes both deprecated names and sets the corresponding boolean flag:

  ```java
  else if (name.id.equals(JIT_COMPILE) || name.id.equals(EXPERIMENTAL_COMPILE))
      this.jitCompileParamExists = true;
  ```

  ```java
  else if (name.id.equals(REDUCE_RETRACING) || name.id.equals(EXPERIMENTAL_RELAX_SHAPES))
      this.reduceRetracingParamExists = true;
  ```

So the *detection* half of #139 — what was actually missing in 2023 — is now in place.

The remaining piece this PR proposed (adding `hasExperimentalCompileParam()` / `hasExperimentalRelaxShapesParam()` accessors) would be **pure aliases** to the existing `isJitCompileParamExists()` / `isReduceRetracingParamExists()` accessors, since both deprecated and modern names share the same internal flag. The accessors don't carry any new information; they're cosmetic API additions.

Two further obstacles to revival:

1. **Naming convention drift**: the PR uses `hasXxxParam()`, but the current code's convention is `isXxxParamExists()` (verifiable via `grep -nE "public boolean is.*Param" Function.java`). A revival would need to rename to match.
2. **API-surface burden**: adding aliases that always return the same value as existing methods clutters the API without enabling new use cases.

Recommending closure. If at some point we want to *separate* the booleans (track "user wrote `experimental_compile`" vs. "user wrote `jit_compile`" distinctly), that's a different change with real semantic content — file as a fresh issue if needed.

PR #144, opened 2023-02-10
Extract tf.function decorator keyword arguments

### LINKED ISSUE #136: Extract tf.function decorator arguments
Currently, we are not dealing with `tf.function` decorator parameter **values**, just with the **existence** of those parameters.

Do we need dataflow or does parsing literals enough? How often are variables used to decorator arguments? Can we tell when variables are used? If we go only with literals, can we throw an exception when a variable is used so that we know how many times this happens. Better yet, we can "fail" the function by adding a precondition failure to the function. In other words, if we can't exact the decorator argument, chances are we won't be able to get a good handle on the function's status.

tatianacv:
Fixes #136 .

## Commits

- [x] I slowed down and took a deep breadth before committing. Otherwise, I will start a pristine branch and reissue this pull request.
- [x] Before committing my changes, I checked my working directory carefully. I only staged the changes I wanted and ensured it was of the utmost quality.
- [x] I used `git diff` and `git status` to check my work carefully before committing.
- [x] I completely understand what I am committing and why.
- [x] I completely understand why things are the way they are.
- [x] The change I am proposing makes complete sense to me.
- [x] For the most part, the changes in this pull request correspond to the problem or task I am solving or performing. In other words, only the changes that are supposed to be there are the ones included in this pull request.

## Expectations

Extract `tf.function` decorator **keyword** parameter values. We are obtaining the decorator’s argument literal values. We throw an exception when we find something we cannot process, like for example, non-literal values.

## Testing

- [x] If this is a bug fix, I was able to reproduce the problem locally **before** I made any changes (can use `git stash` to temporarily reset the working directory). 
- [x] When I made (or put back) my changes, the problem I reproduced above was fixed.

I added tests for many possible values that could arise for each `tf.function` arguments. I used the [`tf.function` documentation](https://www.tensorflow.org/versions/r2.9/api_docs/python/tf/function#args) to understand and make tests for different possible values. I also added test to make sure that for when we are not working with literals, we throw an exception. 

## Final Checks

- [x] The changes I am proposing meet the expectations I have described above.
- [x] I have performed a self-review of my code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation if applicable.
- [x] I have added tests that prove my fix is effective or that my feature works if applicable.
- [x] New and existing unit tests pass locally with my changes.


khatchad (changes requested):

khatchad (changes requested):
Getting there.

khatchad (changes requested):
Javadoc comments are outdated. Many of them say return `String` but the methods don't return `String`. Because of this reason, it doesn't make sense to to use types in the comments. The comments are used to explain what the code doesn't tell us. We can see from the method signature what the return type is. The javadoc should explain the concept of what is being returned.

khatchad:
Closing as stale. Linked issue #136 was closed 2023-06-02 (its target was addressed elsewhere). This PR (+1846/-54 across 77 files, dirty against `main` for 2.5 years) is no longer the path. Reopen if a revival of the decorator-argument extraction surface area becomes useful for the input-signature inference work, but as it stands the cost of rebasing 77 files of 2.5-year-old work outweighs reusing it.

PR #463, opened 2026-05-05
Add `Parameter` skeleton + first inferred-tensor-types test

khatchad:
## Summary

Adds a `Parameter` analytical wrapper for per-parameter classification queries that previously lived on `Function`, plus a first end-to-end test exercising `Parameter.getTensorTypes(TensorTypeAnalysis)` against the shape-divergence/same-dtype scenario from `wala/ML`'s `tf2_test_function8.py`. Tests are migrated from the old `argumentsType.args[]` access to the new `List<Parameter>` API.

## What's Here

- **`Parameter`** (`edu.cuny.hunter.hybridize.core.analysis.Parameter`): analytical wrapper around a single positional Python function parameter. Carries `(argumentsType arguments, int index, Function function)` and hosts the per-parameter classification queries:
  - `getIndex()`, `getName()`, `isSelf()`, `getTypeInfo()` (PyDev AST-derived type hint, nullable).
  - `hasTensorTypeHint(IProgressMonitor)`: walks the type-hint AST node, resolves FQNs, returns true on the first tensor-type match.
  - `getTensorTypes(TensorTypeAnalysis)`: queries Ariadne's analysis for the parameter's inferred tensor types. Takes the analysis as an explicit parameter (no stashed state on `Function`).
  - `hasTensorContainer(TensorTypeAnalysis, CallGraph, PythonSSAPropagationCallGraphBuilder, IProgressMonitor)`: container-of-tensors check.
- **`Function`** loses everything that existed only for `Parameter`: the `tensorTypeAnalysis` back-channel field, `getTensorTypeAnalysis()`, `nodeIsTensorTypeHint`, `getAllAttributes`, `TF_TENSOR_FQN`, and `SELF_PARAMETER_NAME`. `inferTensorParameters` iterates `List<Parameter>` and delegates per-parameter classification to `Parameter`. `inferTensorTensorParameters` is renamed to `inferTensorParameters` (the duplicated `Tensor` was a typo, merged from `main` via #464).
- **`testInferredTensorTypes`**: new JUnit test asserting parameter `t` of `func` is inferred to two `TensorType`s with `(DType.FLOAT32, [2,1])` and `(DType.FLOAT32, [2])` shapes. Fixture ported from `wala/ML`'s `tf2_test_function8.py`. Uses Ariadne's `TensorFlowTypes.DType` enum directly (publicly exposed in the same package as `TensorType`); see #470 for the planned migration off the `DType.valueOf(...toUpperCase())` round-trip once `wala/ML#519`'s `TensorType.getDType()` accessor lands.
- **Test-callsite migration**: every existing test site that read `params.args[i]` is migrated to `params.get(i).getName()` / `params.get(i).getTypeInfo()`. Pre-existing tests continue to assert the same things; the old `argumentsType`-shaped surface is gone from test code.

## Cross-Refs

- #465: extending `Parameter` to cover `kwonlyargs`, `vararg`, `kwarg` (out of scope here; this PR ships positional-only).
- #470: bump Ariadne to `0.43.0` and migrate `DType.valueOf(...)` to `tensorType.getDType()` once `wala/ML#519` is consumable.
- #464: rename `inferTensorTensorParameters` to `inferTensorParameters` (merged into `main`; folded here via the merge commit).
- #467: Static-import sweep for `Boolean.TRUE`/`Boolean.FALSE` (merged into `main`; this branch picks it up).

## Test Plan

- [x] `mvn install -DskipTests=true` green locally on the touched modules.
- [x] `testInferredTensorTypes` passes (Ariadne's `testFunction8` covers this exact scenario upstream).
- [x] Existing parameter-related tests (`testHasLikelyTensorParameter*`, `testComputeParameters*`) pass against the migrated `List<Parameter>` surface.
- [ ] Full Tycho CI green.


edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1563,6 +1571,54 @@ public void inferPythonSideEffects(Map<CGNode, OrdinalSet<PointerKey>> mod, Call
...
+	 * per entry.
+	 * <p>
+	 * This is the input-signature-inference layer's analytical entry point on a per-{@link Function} basis. Call once per function after
+	 * {@link #inferTensorTensorParameters} (the rolled-up boolean predicate {@link #getHasTensorParameter} continues to function
+	 * independently of this).
+	 *
+	 * @param tensorAnalysis Pre-computed Ariadne tensor analysis.
+	 * @param monitor For progress.
+	 */
+	public void computeParameterList(TensorTypeAnalysis tensorAnalysis, IProgressMonitor monitor) {
```
khatchad: I'm unsure if I understand the existence of this method. It seems to be modeled after `computeHybridization` (public method, progress monitor, shared prefix), however, that method exists because determining whether a function is hybrid seems to be computationally expensive. Thus, we pulled this code out of the constructor, where it originally was, and put it in this method. That way, one may create a `Function` rather cheaply and then compute its computationally expensive properties later. However, the parameter list doesn't seem to share this characteristic, and I wonder if its code should be in the constructor instead.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1814,6 +1870,43 @@ public void setRefactoring(Refactoring refactoring) {
...
+	 * {@code paramExpr}. Mirrors the iteration in {@link #tensorAnalysisIncludesParameter} but returns the actual tensor-type set rather
+	 * than a presence boolean. Used by {@link #computeParameterList} to populate {@link Parameter#getInferredTensorTypes}.
+	 *
+	 * @param analysis The {@link TensorTypeAnalysis} produced by Ariadne.
+	 * @param paramExpr The Jython parameter name expression.
+	 * @param paramName The textual parameter name.
+	 * @param monitor For progress.
+	 * @return Possibly-empty unmodifiable set of inferred tensor types.
+	 */
+	private Set<TensorType> getInferredTensorTypesForParameter(TensorTypeAnalysis analysis, exprType paramExpr, String paramName,
```
khatchad: Why is this code not in the `Parameter` class? Seems related to `Parameter`s.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
```diff
@@ -0,0 +1,72 @@
...
+ * {@link Parameter} is constructed with the relevant per-parameter slice in hand and exposes plain accessors. There is no Ariadne
+ * re-invocation per {@link Parameter}.
+ */
+public final class Parameter {
+
+	private final ParameterDefinition definition;
+
+	private final Set<TensorType> inferredTensorTypes;
+
+	public Parameter(ParameterDefinition definition, Set<TensorType> inferredTensorTypes) {
```
khatchad: This should be `public Parameter(ParameterDefinition definition)` and the tensor types should be computed internally.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
```diff
@@ -0,0 +1,72 @@
...
+ * Unlike {@link Function}, {@link Parameter} does not lazily compute its analysis state. Ariadne's {@code TensorTypeAnalysis} is built once
+ * per function at the {@link Function} layer, and the per-parameter results are queried out of that already-built analysis. So a
+ * {@link Parameter} is constructed with the relevant per-parameter slice in hand and exposes plain accessors. There is no Ariadne
+ * re-invocation per {@link Parameter}.
+ */
+public final class Parameter {
+
+	private final ParameterDefinition definition;
+
+	private final Set<TensorType> inferredTensorTypes;
```
khatchad:
```suggestion
	private final Set<TensorType> tensorTypes;
```

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/ParameterDefinition.java
```diff
@@ -0,0 +1,116 @@
...
+	 */
+	public exprType getNameExpr() {
+		return this.arguments.args[this.index];
+	}
+
+	public String getName() {
+		return NodeUtils.getRepresentationString(this.getNameExpr());
+	}
+
+	public File getContainingFile() {
```
khatchad: We don't need this. We can just return the `Function`.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
```diff
@@ -0,0 +1,188 @@
...
+	 * current {@link TensorTypeAnalysis#iterator()} contract, "tensor with unknown types" (i.e. a {@code TensorVariable} with empty state)
+	 * and "not a tensor" (no {@code TensorVariable} bound to the matching pointer key) are indistinguishable, so an empty result means one
+	 * of those two cases without telling them apart. Honoring the wala/ML lattice distinction would require a richer Ariadne-side query;
+	 * left for a future enhancement.
+	 *
+	 * @return Unmodifiable, possibly-empty set of inferred tensor types. Never {@code null}.
+	 * @throws IllegalStateException If the owning {@link Function} has no recorded {@link TensorTypeAnalysis} yet.
+	 */
+	public Set<TensorType> getTensorTypes() {
+		TensorTypeAnalysis analysis = this.function.getTensorTypeAnalysis();
```
khatchad: This is a poor design. Why not pass `TensorTypeAnalysis` as a parameter?

khatchad: Agreed and addressed in e04f24d. `Parameter.getTensorTypes(TensorTypeAnalysis)` now takes the analysis as a parameter, and the back-channel (`Function.tensorTypeAnalysis` field + `Function.getTensorTypeAnalysis()` accessor + the `IllegalStateException` precondition) is gone. Test plumbing updated to surface the per-project analysis through the test helper.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
khatchad: Superseded by e04f24d. Went with method-parameter (`getTensorTypes(TensorTypeAnalysis)`) per the later thread (https://github.com/ponder-lab/Hybridize-Functions-Refactoring/pull/463#discussion_r3221472216), so `Parameter` doesn't accept a `ParameterDefinition` and tensor types aren't computed at construction. Leaving this note so the internal-storage shape isn't re-proposed without seeing the prior consideration.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
khatchad: Superseded by e04f24d. Caching tensor types as a `final` field at construction would require `Parameter` to know the `TensorTypeAnalysis` at construction time, but `Parameter`s are built in `Function`'s constructor before the analysis exists. The landed design takes the analysis at query time via `getTensorTypes(TensorTypeAnalysis)` instead.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1575,91 +1594,81 @@ public void inferTensorTensorParameters(TensorTypeAnalysis tensorAnalysis, CallG
...
-						selfParam = true;
-						continue; // next parameter.
-					}
-
-					// check a special case where we consider type hints.
-					boolean followTypeHints = this.getAlwaysFollowTypeHints() || this.getHybridizationParameters() != null
-							// TODO: Actually get the value here (#111).
-							&& this.getHybridizationParameters().isExperimentalFollowTypeHintsParamExists();
+			// don't consider `self` as a tensor.
+			if (paramInx == 0 && paramName.equals(SELF_PARAMETER_NAME)) {
```
khatchad:
```suggestion
			if (paramInx == 0 && param.isSelf())) {
```

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1731,63 +1740,8 @@ public boolean isHybridizationAvailable() {
...
-					// we are not checking the parameters' expression length because Ariadne includes the type hint in the length while
-					// PyDev does not.
-					return lhsBeginColumn == rhsBeginColumn && lhsBeginLine == rhsBeginLine;
-				}
-			}
-		}
-
-		return false;
+		List<Parameter> parameters = this.getParameters();
+		return parameters.size() > 1 && parameters.get(0).getName().equals(SELF_PARAMETER_NAME);
```
khatchad:
```suggestion
		return parameters.size() > 1 && parameters.get(0).isSelf();
```

edu.cuny.hunter.hybridize.tests/resources/HybridizeFunction/testInferredTensorTypes/in/A.py
```diff
@@ -0,0 +1,24 @@
...
+
+n = random()
+
+if n > 0.5:
+    l = [[1.0], [3.0]]
+else:
+    l = [1.0, 3.0]
+
+a = tf.constant(l)
+assert a.shape == (2, 1) or a.shape == (2,)
```
khatchad, line 23: Missing assertion for `dtype`.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -1879,6 +1892,53 @@ public void testHasLikelyTensorParameter4() throws Exception {
...
+
+		Set<TensorType> inferred = t.getTensorTypes(this.lastTensorTypeAnalysis);
+		assertNotNull(inferred);
+		assertEquals("Two tensor types should be inferred (shape divergence, same dtype).", 2, inferred.size());
+
+		// Both types must be float32 (same dtype).
+		for (TensorType tt : inferred)
+			assertEquals("float32", tt.getCellType());
+
+		// The shapes should be {(2, 1), (2,)} — collapse to lengths to compare without depending on Dimension equality semantics.
```
khatchad: What about `dtype`?

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
khatchad: Done in 1752dc8. Added `Parameter.isSelf()` (and moved the `"self"` constant onto `Parameter`); this call site is now `if (paramInx == 0 && param.isSelf())`. Follow-up 30f1898 also drops the redundant `paramInx == 0` after pulling the position check into `isSelf()` itself.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
khatchad: Done in 1752dc8. `isMethod()` now reads `parameters.size() > 1 && parameters.get(0).isSelf()`.

edu.cuny.hunter.hybridize.tests/resources/HybridizeFunction/testInferredTensorTypes/in/A.py
khatchad: Done in 1752dc8. Added `assert a.dtype == tf.float32` ahead of the shape assertion so the runtime fixture mirrors the JUnit-side expectations.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
khatchad: Done in 1752dc8. The separate dtype-per-element and shape-set checks are now folded into a single `Set<Map.Entry<String, List<Integer>>>` assertion comparing against the expected `{(float32, (2,1)), (float32, (2,))}` set. Each inferred type's dtype is now pinned to its shape rather than checked independently.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1575,91 +1592,81 @@ public void inferTensorTensorParameters(TensorTypeAnalysis tensorAnalysis, CallG
...
-						// try to get its type from the AST.
-						TypeInfo argTypeInfo = NodeUtils.getTypeForParameterFromAST(paramName, functionDef);
+			// if we are considering type hints.
+			if (followTypeHints) {
+				LOG.info("Following type hints for: " + this + " and parameter: " + paramName + ".");
 
-						if (argTypeInfo != null) {
-							LOG.info("Found type for parameter " + paramName + " in " + this + ": " + argTypeInfo.getActTok() + ".");
+				// try to get its type from the AST.
+				TypeInfo argTypeInfo = NodeUtils.getTypeForParameterFromAST(paramName, functionDef);
```
khatchad: This looks like something that should be internal to the `Parameter` class.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1575,91 +1592,81 @@ public void inferTensorTensorParameters(TensorTypeAnalysis tensorAnalysis, CallG
...
-								this.hasTensorParameter = Boolean.TRUE;
-								LOG.info(this + " likely has a tensor parameter: " + paramName + " due to a type hint.");
-								monitor.worked(1);
-								this.addInfo(TYPE_INFERENCING, "Used a type hint to infer tensor type for parameter: " + paramName + ".");
-								continue; // next parameter.
-							}
-						}
-					}
+					exprType node = argTypeInfo.getNode();
+					Set<Attribute> allAttributes = getAllAttributes(node);
```
khatchad: Is there any other code that should be made internal to `Parameter`?

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1575,91 +1592,81 @@ public void inferTensorTensorParameters(TensorTypeAnalysis tensorAnalysis, CallG
...
 					monitor.worked(1);
+					continue; // next parameter.
 				}
 
-				// True if there is only one parameter that is self.
-				final boolean onlySelfParam = actualParams.length == 1 && selfParam;
+				// Check for containers of tensors.
+				if (this.tensorAnalysisIncludesParameterContainer(tensorAnalysis, paramInx, callGraph, builder,
+						monitor.slice(IProgressMonitor.UNKNOWN))) {
+					this.hasTensorParameter = Boolean.TRUE;
```
khatchad: We should use a static import here for `Boolean.TRUE`, but that's out-of-scope for this PR. We should start a new PR against `main` for that.

khatchad: Filed as #467, sweeping `Boolean.TRUE`/`Boolean.FALSE` across the three source files that use them (`HybridizeFunctionRefactoringProcessor`, `Function`, `HybridizeFunctionRefactoringTest`).

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
khatchad: Three other candidates I see in `inferTensorParameters`'s per-parameter loop:

1. **The whole `followTypeHints` block (lines 1614 to 1634):** encapsulate as `Parameter.hasTensorTypeHint(IProgressMonitor)` returning a boolean. Couples `Parameter` to PyDev's `Attribute`, `IProgressMonitor`, and `Function.attributesHaveTensorTypeHints` (currently `private`; would need exposing or moving the TF-name knowledge onto `Parameter`).
2. **The tensor-container check at line 1647:** `this.tensorAnalysisIncludesParameterContainer(tensorAnalysis, paramInx, callGraph, builder, monitor.slice(...))` could become `param.hasTensorContainer(tensorAnalysis, callGraph, builder, monitor)`. Wide parameter list (four objects threaded through), but the method answers a per-parameter question.
3. **The `paramName` and `paramInx` locals (lines 1599, 1600):** already covered by `param.getName()`/`param.getIndex()`; keeping them is fine, dropping them inlines the access at each use.

The general direction "every per-parameter classification lives on `Parameter`" matches the existing surface (`isSelf`, `getTensorTypes`, `getTypeInfo`). My read: (1) is the clearest win and groups type-hint inference under one name; (2) is a clean encapsulation but widens `Parameter`'s parameter surface; (3) is purely stylistic.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
khatchad: Done in a2fe629. `Parameter.getTypeInfo()` returns the AST-derived `TypeInfo` for this parameter, and the call site in `inferTensorParameters` becomes `TypeInfo argTypeInfo = param.getTypeInfo()`. The `functionDef` local that was only there to feed `getTypeForParameterFromAST` also drops out.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1575,91 +1521,67 @@ public void inferTensorParameters(TensorTypeAnalysis tensorAnalysis, CallGraph c
...
-							Set<Attribute> allAttributes = getAllAttributes(node);
+			// If this function is in the call graph.
+			if (!nodes.isEmpty()) {
+				// Ask the parameter directly: does Ariadne associate any tensor type with it?
+				if (!param.getTensorTypes(tensorAnalysis).isEmpty()) {
+					this.hasTensorParameter = TRUE;
+					LOG.info(this + " likely has a tensor parameter: " + param.getName() + " due to tensor analysis.");
+					monitor.worked(1);
+					continue; // next parameter.
+				}
```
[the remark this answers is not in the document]
khatchad: Declining. Verified against Ariadne 0.43.0's `TensorTypeAnalysis.initializeVariables` (`com.ibm.wala.cast.python.ml/source/com/ibm/wala/cast/python/ml/analysis/TensorTypeAnalysis.java:568`):

```java
} else {
    getOut(src).state = null; // unknown tensor — distinguishable from empty (not-a-tensor)
}
```

The seeding model now has three lattice states:

1. `state == null`: unknown tensor (is a tensor, types unknown).
2. `state == empty`: not a tensor (got seeded, no types inferred).
3. `state == non-empty`: known tensor with those types.

The previous `tensorAnalysisIncludesParameter` predated this distinction: at the time, anything seeded was a tensor, so "matched + non-null `TensorVariable`" correctly meant "has tensor parameter". With the lattice, "matched + empty-state" now means "got seeded but isn't a tensor", which is exactly what the new `inferTensorParameters` excludes via `!getTensorTypes(...).isEmpty()`. The `Parameter.getTensorTypes` Javadoc documents that empty-set return covers both "no entries" and "matched-but-empty-state" cases.

edu.cuny.hunter.hybridize.tests/META-INF/MANIFEST.MF
```diff
@@ -23,6 +23,8 @@ Require-Bundle: edu.cuny.citytech.refactoring.common.tests;bundle-version="5.2.0
 Export-Package: edu.cuny.hunter.hybridize.tests
 Import-Package: com.ibm.wala.cast.python.client;version="0.1.0",
  com.ibm.wala.cast.python.loader;version="0.1.0",
+ com.ibm.wala.cast.python.ml.analysis;version="0.29.0",
+ com.ibm.wala.cast.python.ml.types;version="0.29.0",
```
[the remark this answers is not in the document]
khatchad: Fixed in 7eaa845. The core bundle's `com.ibm.wala.cast.python.ml.analysis` import was bumped from the stale `0.16.0` pin to `0.29.0`, matching the test bundle and every other Ariadne-package import.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -673,6 +684,9 @@ public ResourceAttributes getResourceAttributes() {
 		else
 			assertFalse(status.isOK());
 
+		Map<IProject, TensorTypeAnalysis> analysisMap = processor.getProjectToTensorTypeAnalysis();
+		this.lastTensorTypeAnalysis = analysisMap.isEmpty() ? null : analysisMap.values().iterator().next();
```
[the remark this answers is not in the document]
khatchad: Fixed in 7eaa845. The test now asserts `lastTensorTypeAnalysis != null` before passing it to `Parameter.getTensorTypes(...)`, failing the test cleanly with a directed message rather than NPEing through `requireNonNull`.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
```diff
@@ -0,0 +1,328 @@
...
+ * ({@code argumentsType} parent + positional index + owning {@link Function}) and hosts the per-parameter classification queries that would
+ * otherwise live on {@link Function}: type-hint detection ({@link #hasTensorTypeHint(IProgressMonitor)}), Ariadne tensor-type lookup
+ * ({@link #getTensorTypes(TensorTypeAnalysis)}), and tensor-container detection
+ * ({@link #hasTensorContainer(TensorTypeAnalysis, CallGraph, PythonSSAPropagationCallGraphBuilder, IProgressMonitor)}).
+ * <p>
+ * Intentionally narrow public surface: {@link #getIndex()}, {@link #getName()}, {@link #isSelf()}, {@link #getTypeInfo()},
+ * {@link #hasTensorTypeHint(IProgressMonitor)}, {@link #getTensorTypes(TensorTypeAnalysis)},
+ * {@link #hasTensorContainer(TensorTypeAnalysis, CallGraph, PythonSSAPropagationCallGraphBuilder, IProgressMonitor)}, plus
+ * {@code equals}/{@code hashCode}/{@code toString}. Constructed only by {@link Function} (package-private constructor).
+ */
```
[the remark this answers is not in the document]
khatchad: PR description rewritten to reflect the final design: no `ParameterDefinition` is added (the work consolidated into a single `Parameter` class), and the methods listed (`computeParameterList`, `getInferredTensorTypesForParameter`) never existed in the landed code.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
```diff
@@ -0,0 +1,328 @@
...
+		for (Pair<PointerKey, TensorVariable> pair : analysis) {
+			PointerKey pointerKey = pair.fst;
+			if (pointerKey instanceof LocalPointerKey) {
+				LocalPointerKey localPointerKey = (LocalPointerKey) pointerKey;
+				if (localPointerKey.isParameter() && this.matches(localPointerKey)) {
+					TensorVariable tensorVariable = pair.snd;
+					if (tensorVariable != null)
+						result.addAll(tensorVariable.getTypes());
+				}
+			}
```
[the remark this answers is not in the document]
khatchad: Fixed in 7eaa845. `Parameter.getTensorTypes` now throws `IllegalStateException` when a matching `PointerKey` has a null `TensorVariable`, restoring the invariant the prior `Function.tensorAnalysisIncludesParameter` enforced.

edu.cuny.hunter.hybridize.core/META-INF/MANIFEST.MF
```diff
@@ -28,6 +28,7 @@ Import-Package: com.google.common.collect,
  com.ibm.wala.cast.python.loader;version="0.29.0",
  com.ibm.wala.cast.python.ml.analysis;version="0.16.0",
```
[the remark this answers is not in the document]
khatchad: Fixed in 7eaa845 (same root cause as the test-bundle version-mismatch thread). Core's `.analysis` import is now `0.29.0`.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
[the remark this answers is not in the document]
khatchad: Updated the PR title to drop `ParameterDefinition` — the skeleton landed as just `Parameter` (the `Definition` companion was dropped from the design during review). PR body never named the type, so no other reference to clean up.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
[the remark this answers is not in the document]
khatchad: Addressed in a later commit on this branch: `getTensorTypes` now throws `IllegalStateException("Tensor variable was null even though the matching PointerKey is present.")` when the matching pointer key has a null `TensorVariable`, preserving the previous invariant. Current source: `Parameter.java:261`.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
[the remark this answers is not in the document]
khatchad: Mitigated by the test itself: `testInferredTensorTypes` runs `assertNotNull("Test helper should have captured the per-project tensor analysis.", this.lastTensorTypeAnalysis)` before passing the field into `Parameter.getTensorTypes(...)`, so a null `lastTensorTypeAnalysis` fails the test with a clear message rather than NPEing inside `Parameter`. `Parameter.getTensorTypes` itself declares the param non-null in Javadoc and `Objects.requireNonNull`s the argument.

edu.cuny.hunter.hybridize.tests/META-INF/MANIFEST.MF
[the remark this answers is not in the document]
khatchad: Addressed in a later commit: the core bundle's `com.ibm.wala.cast.python.ml.analysis` import was bumped from `0.16.0` to `0.29.0` in this branch's manifest diff, so the two bundles now agree on the version pin.

edu.cuny.hunter.hybridize.core/META-INF/MANIFEST.MF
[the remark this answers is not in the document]
khatchad: Same fix as the sibling thread on the tests manifest: the pre-existing `ml.analysis;version="0.16.0"` line was bumped to `0.29.0` to match the newly-added `ml.types;version="0.29.0"`. Both imports now pin a single consistent minimum.

Issue #29, opened 2022-08-04
Determine if a project using using TF1 or TF2


khatchad:
I think we're going to need to know this for precondition checking. How to do it? I found this from https://www.tensorflow.org/api_docs/python/tf/Tensor:
> The primary use case for this API is to put tensors in a set/dictionary. We can't put tensors in a set/dictionary as `tensor.__hash__()` is no longer available starting Tensorflow 2.0.

Can we test for the availability of that function? Moreover, I'm not so convinced that the `requirements.txt` would not be useful here if it exists.

Lastly, we may run into a case where one file in a project using TF1 and another TF2. Thus, this may be on a function/file basis.

PR #476, opened 2026-05-12
Pin current behavior on TF1 code

khatchad:
## Summary

Towards #156. This PR adds the pinning test only; once TF1/TF2 detection (#29) lands and the refactoring starts rejecting TF1-style code, the assertion will need to be revisited as separate follow-up work. Adds a regression test (`testTF1Code`) and fixture that exercises the refactoring against pure-TF1 code. The desired safety property—refuse to add `@tf.function` to TF1 code, because the API does not exist there (or has incorrect semantics under `disable_v2_behavior`)—depends on TF1/TF2 detection (#29) landing first.

Until then, this test pins the *current (incorrect)* behavior: the refactoring recommends `CONVERT_TO_HYBRID`. A `TODO` references the issue. When the fix lands, the test fails; the natural response is to invert the assertion to assert the refactoring does *not* contain `CONVERT_TO_HYBRID` and drop the `TODO`.

## Why Land the Test Now

- **Pins the bad behavior.** Future analysis changes that accidentally start doing the right thing for the wrong reason still trip the test.
- **Discoverable forcing function.** When #29 lands, the failing test is the natural place to wire the new precondition into the assertion.
- **CI coverage on the breaking path.** A user-report-only path is fragile; this makes it CI-checked even before correctness lands.

## Fixture

`edu.cuny.hunter.hybridize.tests/resources/HybridizeFunction/testTF1Code/in/A.py`:

```python
import tensorflow.compat.v1 as tf

tf.disable_v2_behavior()


def my_function(x):
    return tf.constant(1)


my_function(tf.constant(2))
```

## Verification

`./mvnw verify` green—471 tests, 0 errors, 0 failures, 11 skipped. The new test passes against current main; it's pinning current behavior.


edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -8093,4 +8093,22 @@ public void testSpeculativeAnalysis4() throws Exception {
...
+
+	/**
+	 * Pinning test for https://github.com/ponder-lab/Hybridize-Functions-Refactoring/issues/156. The fixture is a pure TF1 codebase
+	 * (`import tensorflow.compat.v1 as tf` followed by `tf.disable_v2_behavior()`). The refactoring should refuse to add `@tf.function`
+	 * here, because `tf.function` either does not exist in pure TF1 or has incorrect semantics under `disable_v2_behavior`. That
+	 * precondition cannot be implemented until TF1/TF2 detection (#29) lands. Until then, this test pins the current (incorrect)
+	 * behavior: the refactoring proceeds and recommends `CONVERT_TO_HYBRID`. When #29 + the precondition land, this assertion will
+	 * fail; invert it (assert the refactoring does <i>not</i> contain `CONVERT_TO_HYBRID`) and drop the `TODO`.
+	 *
+	 * TODO: Invert assertion once #29 / #156 fix lands.
```
khatchad: #156 is about adding a test; we won't flip anything just for adding a test.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -8093,4 +8093,21 @@ public void testSpeculativeAnalysis4() throws Exception {
...
 				.filter(e -> e.getCode() == SPECULATIVE_ANALYSIS.getCode()).count());
 	}
+
+	/**
+	 * Pinning test for https://github.com/ponder-lab/Hybridize-Functions-Refactoring/issues/156. The fixture is a pure TF1 codebase
+	 * (`import tensorflow.compat.v1 as tf` followed by `tf.disable_v2_behavior()`). The refactoring should refuse to add `@tf.function`
+	 * here, because `tf.function` either does not exist in pure TF1 or has incorrect semantics under `disable_v2_behavior`. That
+	 * precondition cannot be implemented until TF1/TF2 detection (#29) lands; until then, this test pins the current (incorrect) behavior:
+	 * the refactoring proceeds and recommends `CONVERT_TO_HYBRID`. When the fix lands, this assertion will fail; invert it (assert the
+	 * refactoring does <i>not</i> contain `CONVERT_TO_HYBRID`) and drop the comment below.
```
khatchad: Since this is the wrong behavior we are capturing, we need a `TODO` comment here explaining that the test should change once the issue is solved.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -8093,4 +8093,22 @@ public void testSpeculativeAnalysis4() throws Exception {
...
 				.filter(e -> e.getCode() == SPECULATIVE_ANALYSIS.getCode()).count());
 	}
+
+	/**
+	 * Pinning test for https://github.com/ponder-lab/Hybridize-Functions-Refactoring/issues/156. The fixture is a pure TF1 codebase
+	 * (`import tensorflow.compat.v1 as tf` followed by `tf.disable_v2_behavior()`). The refactoring should refuse to add `@tf.function`
+	 * here, because `tf.function` either does not exist in pure TF1 or has incorrect semantics under `disable_v2_behavior`. That
+	 * precondition cannot be implemented until TF1/TF2 detection (#29) lands. Until then, this test pins the current (incorrect)
+	 * behavior: the refactoring proceeds and recommends `CONVERT_TO_HYBRID`. When #29 + the precondition land, this assertion will
+	 * fail; invert it (assert the refactoring does <i>not</i> contain `CONVERT_TO_HYBRID`) and drop the `TODO`.
```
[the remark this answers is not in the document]
khatchad: Good catch — adopted the rewording in b49a369. The Javadoc now says "TF1-style code (TF2 running with `tf.compat.v1` and `disable_v2_behavior()`)" so the description matches what the fixture actually is.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
khatchad: Right — fixed in b49a369. Dropped the prescriptive "invert it" instruction from the Javadoc and the `// TODO` comment. Both now just say "revisit when the refactoring starts rejecting TF1-style code," so #156 only owns the test-adding work and the eventual update belongs to whatever lands the precondition.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
khatchad: Added in 640d67c (and updated in b49a369). The `// TODO` is now on the first line of the method body — `// TODO: Revisit once TF1/TF2 detection (#29) lands and the refactoring starts rejecting TF1-style code.` Spotless was collapsing it when it sat in a paragraph break inside the Javadoc, so the method-body placement keeps it visible at the assertion-site.

PR #480, opened 2026-05-13
Infer input signature for single-context tensor parameter

khatchad:
## Summary

First step toward input-signature inference. Adds the machinery for the simplest case: a function called once with a single tensor of concrete dtype and shape. The inferred signature is exactly that single (dtype, shape) tuple, wrapped.

`inferInputSignature` dispatches per parameter into one of three categories, named upfront so the rest of the description can refer back without restating:

| Label | Condition on parameter | Behavior |
|---|---|---|
| **(a)** Truly non-tensor | `isTensor() != TRUE`—no classification path fired | Drop signature; emit a per-parameter `INPUT_SIGNATURE_INFERENCE` INFO with a *source-side* recovery suggestion (annotate as `tf.Tensor`, wrap call sites with `tf.constant(...)`, rerun) |
| **(b)** Tensor-classified but no Ariadne shape/dtype | `isTensor() == TRUE` AND `getTensorTypes().isEmpty()`—classification came from a type-hint annotation or a tensor-container detection, not from Ariadne's call-site analysis | Drop signature; emit a per-parameter INFO that references #509—recovery is *tool-side* (extend the `Parameter` API to surface what's available) |
| **(c)** Ariadne call-site hit | `isTensor() == TRUE` AND `getTensorTypes()` non-empty | Reduce the multi-context set via the private `inferSpec` and add the spec to the signature |

The three classification paths inside `Parameter.classifyAsTensor` are labeled Phase 1 (type-hint annotation), Phase 2 (Ariadne call-site analysis), Phase 3 (tensor-container detection). Category (a) is "no Phase fired," category (b) is "Phase 1 or Phase 3 fired but Phase 2 has no data," category (c) is "Phase 2 fired."

## New Surface

```java
public record InputSignature(List<TensorType> parameterTypes) { ... }

// On Function:
public Optional<InputSignature> inferInputSignature();
```

`inferInputSignature` walks the function's parameters, skips `self`, and dispatches per parameter into one of the three categories above. For category (c), `inferSpec` handles the multi-context reduction: dtype consensus across contexts; if any context has unknown rank (`dims == null`) or ranks disagree, emit `TensorType(dtype, null)` (shape-⊤); else for each dimension position, keep the value if all contexts agree on a concrete `NumericDim`, otherwise emit a `SymbolicDim("?")` wildcard.

## Tests

- `testInputSignatureScenario1`: a function called once with a single tensor of concrete dtype and shape. Phase 2 hit, no disagreement; the signature is the singleton `TensorType`.
- `testInputSignatureNonParameterTensor`: closure case—a function whose body uses a module-level tensor that is not a parameter. The signature contains only the parameter's `TensorType`; the closure-captured tensor does not leak in.
- `testInputSignatureShapeUnknown`: singleton-shape-⊤ branch (one `TensorType` with `getDims() == null`). The shape-⊤ emission preserves the dtype: the result is a coarse `TensorType(FLOAT32, null)` signature.
- `testInputSignatureNonTensorParameter` (category (a)): mixed (tensor + non-tensor) parameter list yields `Optional.empty` and emits an `INPUT_SIGNATURE_INFERENCE` INFO citing the non-tensor parameter with the source-side recovery suggestion.
- `testInputSignatureMultipleNonTensorParameters` (category (a) accumulation): two non-tensor parameters; asserts one INFO per blocking parameter (not just the first).
- `testInputSignatureContainerParameter` (category (b), Phase-3 path): parameter classified as tensor-container with no Phase 2 evidence; drops the signature and emits an INFO referencing #509.
- `testInputSignatureTypeHintOnly` (category (b), Phase-1 path): parameter with a `tf.Tensor` type-hint classified via Phase 1 but with a non-tensor call site, so Phase 2 has no evidence; drops the signature and emits an INFO referencing #509.
- `testInputSignaturePerDimWildcard`: same-rank tensors of differing size at position 0 (`(2,)` vs `(3,)`) emit a `TensorType(FLOAT32, [SymbolicDim("?")])`—the position-level wildcard branch of `inferSpec`.
- `testInferredTensorTypes` (multi-context fixture, relocated alongside the other type-inference tests): rank disagreement (shape `(2,1)` vs `(2,)`) emits `TensorType(FLOAT32, null)`—the shape-level wildcard branch. The bottom-of-file test diff is mostly relocation, not new content.

## Fixtures

- `testInputSignatureScenario1/in/A.py`: single-parameter function called once with a single concrete tensor.
- `testInputSignatureNonParameterTensor/in/A.py`: same shape, plus a module-level `tf.constant` closed over by the body.
- `testInputSignatureShapeUnknown/in/A.py`: parameter reached via `tf.keras.Input(shape=json.loads(...))` so Ariadne emits a shape-⊤ `TensorType`.
- `testInputSignatureNonTensorParameter/in/A.py`: two-parameter function `f(t, n)` called with a tensor and an `int` literal; `n` is non-tensor.
- `testInputSignatureMultipleNonTensorParameters/in/A.py`: two-parameter function `f(m, n)` called with two `int` literals; both non-tensor.
- `testInputSignatureContainerParameter/in/A.py`: single-parameter function `f(xs)` called with `f([tf.constant(...)])`; Phase 3 classifies `xs` as a container.
- `testInputSignatureTypeHintOnly/in/A.py`: single-parameter function `f(x: tf.Tensor)` called with `f(5)`; Phase 1 classifies `x` via the type hint, Phase 2 has no evidence.
- `testInputSignaturePerDimWildcard/in/A.py`: single-parameter function called twice with same-rank, differently-sized tensors.


edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1591,6 +1592,73 @@ public void inferTensorParameters(TensorTypeAnalysis tensorAnalysis, CallGraph c
...
+	 * @return The reduced single {@link TensorType}, or {@link Optional#empty} for cases not yet implemented.
+	 */
+	private static Optional<TensorType> inferSpec(Set<TensorType> contexts) {
+		if (contexts.size() == 1)
+			return Optional.of(contexts.iterator().next());
+
+		// TODO: multi-context handling (scenarios 2-7). Subsequent PRs will extend.
+		return Optional.empty();
+	}
+
```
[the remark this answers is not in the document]
khatchad: Fixed in 593fbba. The singleton-context branch now requires both a concrete dtype (`getDType()` is non-null and not `DType.UNKNOWN`) and a concrete shape (`getDims()` is non-null and every `Dimension` is a `NumericDim`). Non-concrete cases that would fit scenarios 5, 6, 7, or 10 now correctly fall through to `Optional.empty` pending the multi-context implementation in subsequent PRs.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -8108,4 +8111,41 @@ public void testSelfOnlyMethod() throws Exception {
...
+		assertNotNull("Test helper should have captured the per-project tensor analysis.", this.lastTensorTypeAnalysis);
+
+		List<Parameter> parameters = function.getParameters();
+		assertEquals("Function `func` has exactly one parameter `t`.", 1, parameters.size());
+		Parameter t = parameters.get(0);
+
+		// Ariadne pin: single tensor type seen for the single call site.
+		Set<TensorType> ariadne = t.getTensorTypes(this.lastTensorTypeAnalysis);
+		assertEquals("One call site yields one TensorType.", 1, ariadne.size());
+		TensorType single = ariadne.iterator().next();
```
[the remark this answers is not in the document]
khatchad: Fixed in 593fbba. The Ariadne pin now asserts the expected dtype (`DType.FLOAT32`) and shape (`[2]`) values, not just the count. `func(tf.constant([1.0, 2.0]))` should yield exactly `(FLOAT32, [2])`; a regression to the wrong dtype or shape will fail the test directly, instead of slipping past a count-only check and being silently reused by `assertSame`.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1591,6 +1593,85 @@ public void inferTensorParameters(TensorTypeAnalysis tensorAnalysis, CallGraph c
...
+	 * Current scope: scenario 1 (single-context: returns the singleton input). All other scenarios return {@link Optional#empty} pending
+	 * implementation in subsequent PRs.
+	 *
+	 * @param contexts The non-empty set of {@link TensorType}s Ariadne associated with the parameter across call contexts.
+	 * @return The reduced single {@link TensorType}, or {@link Optional#empty} for cases not yet implemented.
+	 */
+	private static Optional<TensorType> inferSpec(Set<TensorType> contexts) {
+		if (contexts.size() != 1)
+			// TODO: multi-context handling (scenarios 2-7). Subsequent PRs will extend.
+			return Optional.empty();
```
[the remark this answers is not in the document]
khatchad: Fixed in cb67d89. `testInferredTensorTypes` now also asserts that `function.inferInputSignature(this.lastTensorTypeAnalysis)` returns `Optional.empty()` for the multi-context fixture, with a `TODO` for when the multi-context implementation lands. The unsupported branch is now CI-checked.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1591,6 +1593,85 @@ public void inferTensorParameters(TensorTypeAnalysis tensorAnalysis, CallGraph c
...
+	 *
+	 * @param analysis The {@link TensorTypeAnalysis} to query. Non-null.
+	 * @return The inferred signature, or {@link Optional#empty} if no non-{@code self} parameter has any associated tensor types or if
+	 *         {@link #inferSpec} cannot reduce one of the per-parameter sets.
+	 */
+	public Optional<InputSignature> inferInputSignature(TensorTypeAnalysis analysis) {
+		Objects.requireNonNull(analysis);
+
+		List<Parameter> params = this.getParameters();
+		List<TensorType> perParameter = new ArrayList<>();
```
khatchad: What is `perParameter`?

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1591,6 +1593,85 @@ public void inferTensorParameters(TensorTypeAnalysis tensorAnalysis, CallGraph c
...
+
+		for (Parameter param : params) {
+			// `self` is excluded from the signature per the algorithm's spec (scenario 12).
+			if (param.isSelf())
+				continue;
+
+			Set<TensorType> contexts = param.getTensorTypes(analysis);
+			if (contexts.isEmpty())
+				// Ambiguous per Parameter#getTensorTypes Javadoc: "not a tensor" or "tensor with unknown types". Practical default
+				// pending the Ariadne-side lattice query: treat as "not a tensor" and exclude from the signature.
```
khatchad: Let's lookup the exact contract details in https://github.com/ponder-lab/ML.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1591,6 +1593,85 @@ public void inferTensorParameters(TensorTypeAnalysis tensorAnalysis, CallGraph c
...
+				// Algorithm 2 returned bottom for this parameter; the whole signature collapses.
+				return Optional.empty();
+
+			perParameter.add(spec.get());
+		}
+
+		if (perParameter.isEmpty())
+			return Optional.empty();
+
+		return Optional.of(new InputSignature(perParameter));
```
khatchad: What happens when we have a parameter that isn't a tensor?

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/InputSignature.java
```diff
@@ -0,0 +1,20 @@
...
+
+import com.ibm.wala.cast.python.ml.types.TensorType;
+
+/**
+ * The inferred input signature of a hybridizable function: an ordered tuple of {@link TensorType}s, one per non-{@code self}, non-excluded
+ * parameter, in declaration order. Produced by {@link Function#inferInputSignature}, which implements Algorithm 2 of the approach.
+ * <p>
+ * The list is unmodifiable. The constructor copies its argument defensively.
+ */
+public record InputSignature(List<TensorType> parameterTypes) {
```
khatchad, line 16: What happens when we have a parameter that isn't a tensor?

edu.cuny.hunter.hybridize.tests/resources/HybridizeFunction/testInputSignatureScenario1/in/A.py
```diff
@@ -0,0 +1,12 @@
...
+# single (dtype, shape) tuple, wrapped as a singleton InputSignature.
+
+import tensorflow as tf
+
+
+def func(t):
+    return t + 1
+
+
+func(tf.constant([1.0, 2.0]))
```
khatchad: Do we have a test case for a function that has a non-parameter tensor?

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -1932,6 +1935,14 @@ public void testInferredTensorTypes() throws Exception {
 				Map.entry(DType.FLOAT32, Collections.singletonList(2)));
 		assertEquals("Expected (dtype, shape) pairs {(FLOAT32, (2,1)), (FLOAT32, (2,))}", expected, dtypesAndShapes);
 
+		// Pinning assertion on `Function#inferInputSignature` for the multi-context branch. Scenarios 2-7 are not yet implemented, so
+		// multi-context inputs (this fixture's two `TensorType`s for the same parameter) currently collapse to `Optional.empty`. When
+		// the corresponding scenario PR lands, this assertion will fail; replace it with a positive signature check for the
+		// shape-divergence case (`shape=(None, 1)` per the design memo's scenario 3).
+		// TODO: Replace with a positive `InputSignature` assertion once scenarios 2-7 of Algorithm 2 are implemented.
+		assertFalse("Multi-context input currently collapses to `Optional.empty` pending scenarios 2-7. See `Function#inferSpec`.",
```
khatchad: Unsure if we typically use fences for code elements in messages.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -1932,6 +1935,14 @@ public void testInferredTensorTypes() throws Exception {
...
+				function.inferInputSignature(this.lastTensorTypeAnalysis).isPresent());
```
khatchad: `lastTensorTypeAnalysis` looks like a code smell. The `Function`s should already be populated with the needed field values following `this.getFunctions()`; we shouldn't have to perform additional analysis afterwards. Let's check the existing tests to see if analysis is performed following `this.getFunctions` (I doubt so). Let's try to model the existing tests as closely as possible. We are only adding the specific types here; the existing code already has tensor inference. The changes should be minimal.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -8108,4 +8119,48 @@ public void testSelfOnlyMethod() throws Exception {
...
+				.collect(Collectors.toList());
+		assertEquals("Ariadne should infer shape `[2]` for `tf.constant([1.0, 2.0])`.", List.of(2), shape);
+
+		// Algorithm assertion: the inferred signature wraps that single TensorType, unchanged. TensorType does not override `equals`, so
+		// identity is the right check for the single-context case.
+		Optional<InputSignature> signature = function.inferInputSignature(this.lastTensorTypeAnalysis);
+		assertTrue("Scenario 1 yields a non-empty signature.", signature.isPresent());
+		List<TensorType> specs = signature.get().parameterTypes();
+		assertEquals("Single tensor parameter, single entry in the signature.", 1, specs.size());
+		assertSame("Algorithm 2's output for the single-context case is the same `TensorType` instance Ariadne produced.", single,
```
khatchad: We should be hard-coding the expected input signature. I'm unsure what this is actually testing. Also, if `TensorType` doesn't override `equals` and `hashCode` but it should, we should file an issue on Ariadne for that.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
khatchad: Renamed to `specs` in 276a3ea. It is the list of per-parameter `TensorType` specs that becomes `InputSignature.parameterTypes`.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
khatchad: Done in 276a3ea. The inline comment now cites the class-level Javadoc on `com.ibm.wala.cast.python.ml.client.TensorGenerator` as the source of truth for the lattice (⊤ unknown tensor vs ⊥ not a tensor vs concrete set), and notes that `TensorTypeAnalysis.iterator()` filters to entries with `state != null && !state.isEmpty()`, collapsing ⊤ and ⊥ at the query surface. Until a richer Ariadne-side query exposes the distinction, we treat empty as "not a tensor" and exclude the parameter from the signature.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
khatchad: Documented in 276a3ea on both `Function.inferInputSignature` and the `InputSignature` record Javadoc: parameters the analysis classified as non-tensor (empty `Parameter.getTensorTypes()`) are excluded from `InputSignature.parameterTypes`—the signature is tensor-only by definition. If every parameter is non-tensor, the function has no signature to infer and `inferInputSignature` returns `Optional.empty` rather than producing an empty `InputSignature`.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/InputSignature.java
khatchad: Documented in 276a3ea on the `InputSignature` record Javadoc—non-tensor parameters are excluded from `parameterTypes`; an all-non-tensor function yields `Optional.empty` from `Function.inferInputSignature`.

edu.cuny.hunter.hybridize.tests/resources/HybridizeFunction/testInputSignatureScenario1/in/A.py
khatchad: Added in 276a3ea: `testInputSignatureNonParameterTensor` + fixture. The fixture defines a module-level `bias = tf.constant([10.0, 20.0])` closed over by a one-parameter function `func(t)`; the test asserts the inferred signature contains exactly one spec (for `t`) and that the closure-captured tensor does not leak into `InputSignature.parameterTypes`.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
khatchad: Dropped in 276a3ea. JUnit assertion messages now use plain prose (e.g., "One call site yields one TensorType."), no backticks around identifiers.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
khatchad: Refactored in 276a3ea. The `lastTensorTypeAnalysis` test field is gone. `Parameter` now caches its own `Set<TensorType>` (the result, not the input)—populated during `Function.inferTensorParameters` in the existing per-parameter loop, exposed via no-arg `Parameter.getTensorTypes()`. `Function.inferInputSignature()` is correspondingly parameterless. Mirrors the `Function.getHasTensorParameter` no-arg pattern: `Function`s returned from `getFunctions()` are already populated with the needed analysis state, callers do not re-thread the analysis through getters.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
khatchad: Done in 276a3ea. `testInputSignatureScenario1` now hard-codes the expected dtype (`FLOAT32`) and shape (`[2]`) and compares field-by-field against both the pinned Ariadne `TensorType` and the inferred-signature spec.

Field-by-field is the workaround for `TensorType` lacking `equals`/`hashCode` overrides upstream—`assertEquals(expected, actual)` falls back to identity and never passes for two analyzer-constructed instances. Filed wala/ML#532 to add the overrides; once that lands and Ariadne is bumped here, the test simplifies to a single `assertEquals` per spec.

khatchad: Retracting my prior reply. I was wrong: `TensorType` already overrides `equals` and `hashCode` on `(cellType, dims)` (and `Dimension` does too)—has since ponder-lab/ML#156, present in the 0.43.0 jar Hybridize consumes (verified via `javap`). I filed wala/ML#532 on an unverified premise; you closed it correctly as invalid, with regression guards added in ponder-lab/ML#295.

Test simplified to direct `assertEquals(expected, actual)` on `TensorType` in 87a697c. `asShape` helper dropped.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -1914,23 +1906,20 @@ public void testInferredTensorTypes() throws Exception {
...
-		Set<Map.Entry<DType, List<Integer>>> dtypesAndShapes = inferred.stream()
-				.map(tt -> Map.entry(tt.getDType(),
-						tt.getDims().stream()
-								.map(d -> d instanceof TensorType.NumericDim ? ((TensorType.NumericDim) d).value() : Integer.valueOf(-1))
-								.collect(Collectors.toList())))
-				.collect(toSet());
+		Set<TensorType> expected = Set.of(
+				new TensorType(DType.FLOAT32.name().toLowerCase(), List.of(new TensorType.NumericDim(2), new TensorType.NumericDim(1))),
+				new TensorType(DType.FLOAT32.name().toLowerCase(), List.of(new TensorType.NumericDim(2))));
+		assertEquals(expected, inferred);
```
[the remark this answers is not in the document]
khatchad: Pushing back. `TensorType.equals`/`hashCode` and `Dimension.equals`/`hashCode` are overridden upstream on `(cellType, dims)` and `v` respectively, in place since [`5467c89a`](https://github.com/ponder-lab/ML/commit/5467c89a) (ponder-lab/ML#156). Regression guards live at ponder-lab/ML#295 (`TensorTypeTest`). Present in the 0.43.0 jar Hybridize consumes (verified via `javap`).

The "existing test logic avoided this" framing reflects an earlier iteration of mine that hand-rolled a field-by-field comparison on the same misreading—retracted in [#discussion_r3236551225](https://github.com/ponder-lab/Hybridize-Functions-Refactoring/pull/480#discussion_r3236551225). Direct `assertEquals(expected, actual)` is correct.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1551,8 +1553,12 @@ public void inferTensorParameters(TensorTypeAnalysis tensorAnalysis, CallGraph c
 
 			// If this function is in the call graph.
 			if (!nodes.isEmpty()) {
-				// Ask the parameter directly: does Ariadne associate any tensor type with it?
-				if (!param.getTensorTypes(tensorAnalysis).isEmpty()) {
+				// Ask the parameter directly: does Ariadne associate any tensor type with it? Cache the result on the parameter so
+				// subsequent reads via `Parameter#getTensorTypes()` (no-arg) do not need to re-thread the analysis.
+				Set<TensorType> tensorTypes = param.getTensorTypes(tensorAnalysis);
+				param.setTensorTypes(tensorTypes);
```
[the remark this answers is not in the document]
khatchad: Good catch. Fixed in 131529a by hoisting `param.setTensorTypes(param.getTensorTypes(tensorAnalysis))` above the `followTypeHints` shortcut, gated on `!nodes.isEmpty()`. The cache is now populated consistently regardless of which branch fires, so `inferInputSignature` sees the Ariadne-inferred set even for type-hint parameters.

edu.cuny.hunter.hybridize.tests/resources/HybridizeFunction/testInputSignatureScenario1/in/A.py
```diff
@@ -0,0 +1,12 @@
...
+# concrete dtype and shape. The inferred signature is exactly that single (dtype, shape) tuple,
+# wrapped as a singleton InputSignature.
+
+import tensorflow as tf
+
+
+def func(t):
+    return t + 1
+
+
```
khatchad, line 11: Need asserts here.

khatchad: Done in 0f23f79. The fixture now binds the tensor to `x`, asserts `x.dtype == tf.float32` and `x.shape == (2,)`, then invokes `func(x)`. Mirrors the JUnit expectations per the test-fixture three-check protocol.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1591,6 +1601,88 @@ public void inferTensorParameters(TensorTypeAnalysis tensorAnalysis, CallGraph c
...
+
+		// The single-context case requires both a concrete dtype and a concrete shape (non-null dims list, every dim a `NumericDim`).
+		// Non-concrete cases return `Optional.empty` pending implementation in subsequent PRs.
+		if (single.getDType() == null || single.getDType() == DType.UNKNOWN)
+			return Optional.empty();
+		if (single.getDims() == null)
+			return Optional.empty();
+		for (TensorType.Dimension<?> dim : single.getDims())
+			if (!(dim instanceof TensorType.NumericDim))
+				return Optional.empty();
```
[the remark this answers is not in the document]
khatchad: Good catch. Added `testInputSignatureShapeUnknown` in 0f23f79 covering `inferSpec`'s `single.getDims() == null` branch via a fixture that produces shape-⊤ on a concrete dtype: `tf.keras.Input(shape=json.loads("[32]"))`. Ariadne can't trace through `json.loads`, so the parameter's `TensorType` is `(FLOAT32, null)`. Test pins (a) the parameter has a non-null TensorType set (so we're in the singleton branch, not the no-iterator-entry path) and (b) `inferInputSignature` returns `Optional.empty` because of the null-dims check.

The dtype-⊤ (`single.getDType() == DType.UNKNOWN`) and non-`NumericDim` branches remain uncovered—those need different fixtures (`tf.py_function` with no `Tout` spec for dtype-⊤; dynamic-rank-only shape for non-numeric dim). Filing as follow-up since each needs Ariadne probing to confirm a stable repro.

edu.cuny.hunter.hybridize.tests/resources/HybridizeFunction/testInputSignatureNonParameterTensor/in/A.py
```diff
@@ -0,0 +1,14 @@
+# The function `func` closes over `bias`, a module-level tensor. The single parameter `t` is also
+# a tensor (provided at the call site). The inferred input signature should contain only `t`'s
+# tensor type; `bias` is not a parameter and must not leak into the signature.
+
+import tensorflow as tf
+
+bias = tf.constant([10.0, 20.0])
```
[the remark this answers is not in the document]
khatchad: Done in 0f23f79. `bias` is now a `(3, 2)` tensor (vs. `t`'s `(2,)`); the body uses `tf.reduce_sum(bias)` to keep the broadcast valid while preserving the closure capture. An implementation that collected all tensors and deduplicated by `TensorType` would now surface both `(3, 2)` and `(2,)` shapes, failing the singleton assertion on `signature.get().parameterTypes()`.

edu.cuny.hunter.hybridize.tests/resources/HybridizeFunction/testInputSignatureShapeUnknown/in/A.py
```diff
@@ -0,0 +1,18 @@
+# Singleton-non-concrete coverage: the parameter receives a tensor whose dtype Ariadne resolves
+# (float32, the `tf.keras.Input` default) but whose shape it cannot trace through `json.loads`.
+# `inferSpec`'s `single.getDims() == null` branch fires and `inferInputSignature` returns
+# `Optional.empty`.
+
+import json
+
+import tensorflow as tf
```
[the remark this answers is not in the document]
khatchad: > This new fixture directory does not include an `in/requirements.txt`

Added in 36b69cfd with `tensorflow==2.9.3` to match peer fixtures (`testInputSignatureScenario1`, `testInputSignatureNonParameterTensor`).

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -8030,48 +8078,74 @@ public void testSelfOnlyMethod() throws Exception {
...
+
+		List<Parameter> parameters = function.getParameters();
+		assertEquals(1, parameters.size());
+		Parameter t = parameters.get(0);
+
+		Set<TensorType> ariadne = t.getTensorTypes();
+		assertFalse("Expected a non-empty TensorType set so we exercise the singleton branch, not the no-iterator-entry one.",
+				ariadne.isEmpty());
+		assertTrue("Expected a shape-⊤ marker (TensorType with null dims) so `inferSpec`'s shape-⊤ branch fires.",
+				ariadne.stream().anyMatch(tt -> tt.getDims() == null));
```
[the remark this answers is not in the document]
khatchad: > This assertion does not actually pin the singleton non-concrete branch described by the test

Tightened in 36b69cfd: `assertEquals(1, ariadne.size())` followed by `assertNull(single.getDims())` on the lone element. The test now exercises the `single.getDims() == null` branch specifically, not the `contexts.size() != 1` branch.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1448,6 +1450,87 @@ public void inferTensorParameters(TensorTypeAnalysis tensorAnalysis, CallGraph c
...
+	public Optional<InputSignature> inferInputSignature() {
+		List<TensorType> specs = new ArrayList<>();
+
+		for (Parameter param : this.getParameters()) {
+			// `self` is excluded from the signature.
+			if (param.isSelf())
+				continue;
+
+			Set<TensorType> contexts = param.getTensorTypes();
+			if (contexts.isEmpty())
```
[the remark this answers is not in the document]
khatchad: > `inferInputSignature()` skips any parameter whose cached tensor-type set is empty, but `Parameter.isTensorTyped` can classify a parameter as tensor-typed from a type hint before calling `inferTensorTypes`

Fixed upstream by #496, which hoists `inferTensorTypes` to fire unconditionally on `classifyAsTensor` entry (right after the `self` check, gated on `!nodes.isEmpty()`). Post-#496 the Phase 2 cache is populated *before* Phase 1's type-hint shortcut returns, so `getTensorTypes()` no longer returns the empty default for type-hint-classified parameters. This merge brings the hoist onto this branch. The silent-drop scenario described here no longer reproduces.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/InputSignature.java
khatchad: > What happens when we have a parameter that isn't a tensor?

Silently skipped via `continue` in `Function.inferInputSignature` (the `if (contexts.isEmpty())` branch). The signature contains only tensor parameters, in declaration order; non-tensor parameters are omitted.

Two consequences worth flagging:

1. **Positional ambiguity.** A function `def f(t: Tensor, n: int)` yields `InputSignature([TensorType(...)])` — the caller cannot reconstruct that the `TensorType` corresponds to position 0 and not position 1. `tf.function(input_signature=...)` requires positional matching, so the current representation doesn't drive that consumer for mixed-parameter functions.
2. **Mismatch with documented design intent.** The class Javadoc says "the signature is by definition tensor-only" — that's the *current* contract, but it leaves consumers without a way to wire `tf.function(input_signature=...)` for mixed-parameter functions.

Filed #508 to track the representation decision (silent skip vs positional tombstone vs whole-signature rejection). Not in scope for #480.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1448,6 +1451,118 @@ public void inferTensorParameters(TensorTypeAnalysis tensorAnalysis, CallGraph c
...
+	 * <p>
+	 * Current scope: a single-context input with concrete dtype and concrete shape returns the singleton unchanged. Multi-context (#507)
+	 * and other non-concrete cases (#494) return {@link Optional#empty} pending future PRs.
+	 *
+	 * @param contexts The non-empty set of {@link TensorType}s Ariadne associated with the parameter across call contexts.
+	 * @return The reduced single {@link TensorType}, or {@link Optional#empty} for cases not yet implemented.
+	 */
+	private static Optional<TensorType> inferSpec(Set<TensorType> contexts) {
+		if (contexts.size() != 1)
+			// TODO(#507): multi-context handling.
```
khatchad: Why is #507 not in scope of this PR? Only handling one context item here instead of a loop seems odd.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1448,6 +1451,118 @@ public void inferTensorParameters(TensorTypeAnalysis tensorAnalysis, CallGraph c
...
+
+		// The single-context case requires both a concrete dtype and a concrete shape (non-null dims list, every dim a `NumericDim`).
+		// Non-concrete cases return `Optional.empty`; branch coverage tracked at #494. The `!NumericDim` short-circuit is conservative
+		// pending that coverage—a SymbolicDim (e.g., dynamic batch axis) could in principle map to a `None`-position in the emitted
+		// TensorSpec, but until #494's per-branch test pins that mapping, dropping is the safe default.
+		if (single.getDType() == null || single.getDType() == DType.UNKNOWN)
+			return Optional.empty();
+		if (single.getDims() == null)
+			return Optional.empty();
+		if (single.getDims().stream().anyMatch(d -> !(d instanceof TensorType.NumericDim)))
```
khatchad: Use `NumericDim` instead of `TensorType.NumericDim` (use import statement).

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
```diff
@@ -91,9 +92,11 @@ public final class Parameter {
 	private final int index;
 
 	/**
-	 * The possible {@link TensorType}s of this {@link Parameter}.
+	 * The {@link TensorType}s associated with this parameter by the {@link TensorTypeAnalysis}. Populated during the
+	 * {@link Function#inferTensorParameters} pass so {@link #getTensorTypes()} can be called without re-passing the analysis. Empty by
+	 * default, never {@code null}.
 	 */
-	private Set<TensorType> tensorTypes;
+	private Set<TensorType> tensorTypes = Collections.emptySet();
```
khatchad: Why start this as an empty set?

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
```diff
@@ -178,19 +181,6 @@ protected TypeInfo getTypeInfo() {
...
 	}
 
-	/**
-	 * Returns the set of possible {@link TensorType}s for this {@link Parameter}.
-	 *
-	 * @return The set of possible {@link TensorType}s for this {@link Parameter}.
-	 */
-	public Set<TensorType> getTensorTypes() {
-		return this.tensorTypes;
-	}
```
khatchad, line 188: Why removing?

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
```diff
@@ -178,19 +181,6 @@ protected TypeInfo getTypeInfo() {
...
-
-	protected void setTensorTypes(Set<TensorType> tensorTypes) {
-		this.tensorTypes = tensorTypes;
-	}
```
khatchad, line 192: Why removing?

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
```diff
@@ -531,7 +521,33 @@ void inferTensorTypes(TensorTypeAnalysis analysis) {
 			}
 		}
 
-		this.setTensorTypes(unmodifiableSet(result));
+		this.setTensorTypes(result);
```
khatchad: Why this change?

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
```diff
@@ -531,7 +521,33 @@ void inferTensorTypes(TensorTypeAnalysis analysis) {
...
+	 * Ariadne associated no {@link TensorType} with this parameter. Callers that need to distinguish those cases should consult
+	 * {@link #isTensor()} instead—{@code null} means "classifier has not run," {@code FALSE} means "ran and not classified as tensor,"
+	 * {@code TRUE} means "classified as tensor" (in which case {@link #getTensorTypes()} may still be empty if classification came from a
+	 * type hint or container detection rather than Ariadne call-site evidence).
+	 *
+	 * @return Unmodifiable, possibly-empty set of inferred tensor types. Never {@code null}.
+	 */
+	public Set<TensorType> getTensorTypes() {
+		return this.tensorTypes;
+	}
```
khatchad: Why are we moving this down the file?

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
```diff
@@ -531,7 +521,33 @@ void inferTensorTypes(TensorTypeAnalysis analysis) {
...
+
+	/**
+	 * Package-private setter used by {@link Function#inferTensorParameters} to populate the cache read by the no-arg
+	 * {@link #getTensorTypes()}. Wraps the given set in an unmodifiable view at write-time so the no-allocation getter returns an
+	 * already-immutable reference.
+	 */
+	void setTensorTypes(Set<TensorType> tensorTypes) {
+		this.tensorTypes = unmodifiableSet(Objects.requireNonNull(tensorTypes));
```
khatchad: Why are we moving this down the file?

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -1881,6 +1885,51 @@ public void testHasLikelyTensorParameter4() throws Exception {
...
+
+		Parameter t = parameters.get(0);
+		assertEquals("t", t.getName());
+		assertEquals(0, t.getIndex());
+
+		Set<TensorType> inferred = t.getTensorTypes();
+		assertNotNull(inferred);
+
+		Set<TensorType> expected = Set.of(
+				new TensorType(DType.FLOAT32.name().toLowerCase(Locale.ROOT),
```
khatchad: Use the ctor that takes a `DType` directly.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -8031,47 +8080,140 @@ public void testSelfOnlyMethod() throws Exception {
...
-	 * Exercises {@link Parameter#getTensorTypes()} on the shape-divergence/same-dtype scenario ported from wala/ML's
-	 * {@code tf2_test_function8.py}: parameter {@code t} is reached by {@code tf.constant(l)} where {@code l} is one of two literal lists
-	 * of different rank. Ariadne therefore associates two {@link TensorType}s with {@code t}, both {@code float32}, with shapes
-	 * {@code (2, 1)} and {@code (2,)} respectively.
+	 * Input-signature inference for a function called once with a single tensor of concrete dtype and shape. The expected signature is a
+	 * singleton containing that single tensor type.
 	 */
 	@Test
-	public void testInferredTensorTypes() throws Exception {
+	public void testInputSignatureScenario1() throws Exception {
```
khatchad, line 8088: It looks like this method is being changed, but actually it's a new method.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
khatchad: > Why is #507 not in scope of this PR? Only handling one context item here instead of a loop seems odd.

This PR's title and scope is "single-context tensor parameter"—the simplest case. Multi-context reduction has its own design choice in the per-context-disagreement case (union-then-wildcard, strict-drop on disagreement, etc.), so #507 tracks that decision plus its implementation. The pinning assertion in `testInferredTensorTypes` (`assertFalse(...inferInputSignature().isPresent())` with a TODO) flips to a positive `InputSignature` check when #507 lands.

Splitting it out keeps this PR's review focused on the single-context core (the dispatch into the three classification categories, the conservative drop on non-concrete branches, and the INFO suggestion pattern) and avoids tangling that baseline with the multi-context design discussion.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
khatchad: > Use `NumericDim` instead of `TensorType.NumericDim` (use import statement).

Done in a746111d—added `import com.ibm.wala.cast.python.ml.types.TensorType.NumericDim;` and use the unqualified form here.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
khatchad: > Why start this as an empty set?

For the never-null invariant. Without the explicit initialization, `tensorTypes` defaults to `null`, and any code path that reads `getTensorTypes()` before `inferTensorParameters` runs would either NPE on iteration or have to null-guard at every call site. Initializing to `Collections.emptySet()` makes the field non-null from construction, so callers (including the new `Function.inferInputSignature`) can iterate / size-check without a null check.

The setter then preserves the invariant by wrapping with `Objects.requireNonNull(...)`. Together, the `getTensorTypes()` Javadoc can now state "Never `null`" as a hard guarantee—see the updated Javadoc paragraph clarifying that the empty-set return is overloaded between "analysis hasn't run" and "analysis ran and found no `TensorType`," and pointing callers to `isTensor()` for that distinction.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
khatchad: > Why removing?

Not removing—relocated. In the previous push I moved both `getTensorTypes()` and `setTensorTypes()` further down the file (near `inferTensorTypes`, which populates them) and gave them updated Javadoc. That move was stylistic and got pushback in the parallel "why moving this down the file" thread, so a746111d restores them to their original location here. The substantive changes that stay: the new Javadoc explaining the empty-set overload, and the setter's `unmodifiableSet(Objects.requireNonNull(...))` wrap.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
khatchad: > Why removing?

Same as the parallel thread on the getter—not removing, just relocated in the previous push and now restored to the original location in a746111d.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
khatchad: > Why this change?

`setTensorTypes` now wraps the incoming set in `Collections.unmodifiableSet(Objects.requireNonNull(...))` at write-time. Two effects:

1. `getTensorTypes()` becomes allocation-free—the field is already an unmodifiable view, so the getter just returns the reference rather than re-wrapping per call.
2. The setter is single-source-of-truth for the unmodifiability invariant; callers like `inferTensorTypes` no longer need to wrap before passing.

Drops one allocation per `getTensorTypes()` call and removes the previously-redundant `unmodifiableSet(result)` at the `setTensorTypes` callsite.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
khatchad: > Why are we moving this down the file?

Restored to original location in a746111d. The move-down in the previous push was stylistic (grouping the getter near the populating method `inferTensorTypes` rather than near the constructor accessors). Reverted so the diff stays focused on the substantive changes (new Javadoc, unmodifiable-wrap, empty-set default).

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
khatchad: > Why are we moving this down the file?

Same as the parallel thread on the getter—restored to original location in a746111d.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
khatchad: > Use the ctor that takes a `DType` directly.

Done in a746111d—migrated all four call sites in this test file (`testInferredTensorTypes`, `testInputSignatureScenario1`, `testInputSignatureNonParameterTensor`, `testInputSignatureShapeUnknown` already used the typed form via the merge) to `new TensorType(DType.FLOAT32, ...)` with unqualified `new NumericDim(...)`. Dropped the now-unused `java.util.Locale` import.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
khatchad: > It looks like this method is being changed, but actually it's a new method.

Yes—this is a git-diff artifact from the relocation. `testInferredTensorTypes` was moved up to L1894 (alongside other type-inference tests like `testInferredTensorTypesUnderFollowTypeHints`), and `testInputSignatureScenario1` was newly added at the bottom-of-file location where `testInferredTensorTypes` used to live. Git's diff heuristic aligns the two by textual similarity (both start with `Set<Function> functions = this.getFunctions();` followed by a list-of-parameters destructure) and renders the change as a modification of the relocated method rather than a deletion-plus-addition.

The PR body's "Tests" section calls this out: *"`testInferredTensorTypes` (multi-context fixture, relocated alongside the other type-inference tests) ... The bottom-of-file test diff is mostly relocation, not new content."* The clearest read is to look at L1894 (where the relocated `testInferredTensorTypes` now lives) and treat the L8087 hunk as a brand-new `testInputSignatureScenario1` method.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
khatchad: > The lack of a loop here just looks odd. Let's roll #507 in.

Done in e04015c1. `Function.inferSpec` now implements the per-context reduction in full loop form: dtype consensus, rank consensus or shape-⊤, per-position consensus or `SymbolicDim("?")` wildcard. Multi-context with rank disagreement emits a coarse `TensorType(dtype, null)`; per-dim disagreement emits a wildcard at that position. Two existing tests flipped from drop-assertions to coarse-signature-assertions; new `testInputSignaturePerDimWildcard` pins the per-position wildcard branch. Closes #507; narrows #494 to dtype-⊤ only.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
khatchad: We need to move these back to their original position so that we can see the changes being made.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java
khatchad: > Why do we want/need a "never-null invariant? I never asked for it. Let's revert this change.

Reverted in the next push. `tensorTypes` is back to its original declaration (`private Set<TensorType> tensorTypes;` — uninitialized), `setTensorTypes` is back to the simple `this.tensorTypes = tensorTypes;` form, and the call site in `inferTensorTypes` wraps in `unmodifiableSet(result)` as before. The field-default and `requireNonNull` wrap were over-engineering on my part; sorry for the unwanted change.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -1448,6 +1454,139 @@ public void inferTensorParameters(TensorTypeAnalysis tensorAnalysis, CallGraph c
...
+	 * @param contexts The non-empty set of {@link TensorType}s Ariadne associated with the parameter across call contexts.
+	 * @return The reduced single {@link TensorType}, or {@link Optional#empty} for the dtype-⊥ and dtype-⊤ branches.
+	 */
+	private static Optional<TensorType> inferSpec(Set<TensorType> contexts) {
+		// Step 1: dtype consensus. The `|D| ≠ 1 ⇒ ⊥` branch covers both the heterogeneous-dtype case and the empty-set case
+		// (already filtered upstream by `inferInputSignature`'s `contexts.isEmpty()` check).
+		Set<DType> dtypes = contexts.stream().map(TensorType::getDType).collect(Collectors.toSet());
+		if (dtypes.size() != 1)
+			return Optional.empty();
+		DType dtype = dtypes.iterator().next();
```
khatchad: This looks odd and potentially unsafe (e.g., no call to `hasNext()`). Should we have a loop?

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/InputSignature.java
```diff
@@ -0,0 +1,25 @@
...
+ * {@link TensorType}, or {@link Function#inferInputSignature} returns {@link java.util.Optional#empty} instead of producing a partial
+ * signature. The per-parameter dispatch (tensor-classified-with-evidence vs tensor-classified-without-evidence vs truly non-tensor) and its
+ * diagnostics live in {@link Function#inferInputSignature}; see that method's Javadoc and #508 for the rules.
+ * <p>
+ * The list is unmodifiable. The constructor copies its argument defensively.
+ */
+public record InputSignature(List<TensorType> parameterTypes) {
+
+	public InputSignature {
+		Objects.requireNonNull(parameterTypes, "parameterTypes must not be null");
```
khatchad: Why can't `parameterTypes` be null?

khatchad: > Why can't `parameterTypes` be null?

Removed in the next push—the explicit `requireNonNull` was over-defensive (same pattern as the never-null invariant on `Parameter.tensorTypes`). `List.copyOf(...)` still NPEs on null, so the constraint is preserved in spirit without the unrequested explicit check.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/InputSignature.java
```diff
@@ -0,0 +1,25 @@
...
+		parameterTypes = List.copyOf(parameterTypes);
```
khatchad: Why do we need a copy of `parameterTypes`?

khatchad: > Why do we need a copy of `parameterTypes`?

Removed in the next push along with the canonical constructor body—the defensive copy was the same over-engineering pattern as the `requireNonNull` and the never-null invariant on `Parameter.tensorTypes`. The record now stores the passed list directly.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -1881,6 +1885,52 @@ public void testHasLikelyTensorParameter4() throws Exception {
...
+
+		Set<TensorType> expected = Set.of(new TensorType(DType.FLOAT32, List.of(new NumericDim(2), new NumericDim(1))),
+				new TensorType(DType.FLOAT32, List.of(new NumericDim(2))));
+		assertEquals(expected, inferred);
+
+		// Multi-context input with rank disagreement (rank 2 vs rank 1): dtype consensus passes (both float32), shape axis degrades
+		// to ⊥ (null dims). `inferInputSignature` emits a coarse `TensorType(FLOAT32, null)` signature rather than dropping the
+		// parameter.
+		Optional<InputSignature> signature = function.inferInputSignature();
+		assertTrue("Multi-context rank-disagreement emits a coarse `TensorType(FLOAT32, null)` signature.", signature.isPresent());
```
khatchad, line 1922: Does `null` represent a wild-card here?

khatchad: > Does `null` represent a wild-card here?

Yes—`getDims() == null` is the *shape-level* wildcard (entire shape unknown, Ariadne's shape-⊤ representation, equivalent to TF's `TensorSpec(shape=None, dtype=...)` at runtime). It's distinct from the *position-level* wildcard (`SymbolicDim("?")` as an entry in a non-null dims list, equivalent to TF's `TensorSpec(shape=(None, 784), ...)`), which is what `testInputSignaturePerDimWildcard` covers.

The two levels correspond to the two `inferSpec` branches: rank disagreement (or any context with `dims == null`) emits the shape-level wildcard via `TensorType(dtype, null)`; per-dim disagreement with matching ranks emits position-level wildcards via `SymbolicDim("?")` entries in the otherwise-concrete dims list.

Issue #491, opened 2026-05-16
Add full-⊤ (dtype-⊤ and shape-⊤) regression fixture


khatchad:
## Background

Follow-up to #486. The lattice has two orthogonal axes per the wala/ML `com.ibm.wala.cast.python.ml/CONTRIBUTING.md` contract:

- Shape: `null` dims is shape-⊤.
- Dtype: `EnumSet.of(DType.UNKNOWN)` is dtype-⊤.

The regression test added by #486's fix (`testInferredTensorTypesUnknownShapeTop`) exercises shape-⊤ (`tf.keras.Input(shape=json.loads(...))` produces `TensorType(FLOAT32, null)`—dtype is `tf.keras.Input`'s default `float32`). The dtype-⊤ axis is not exercised.

## What's Needed

A fixture where Ariadne genuinely cannot determine the dtype, producing a `TensorType` with `getDType() == DType.UNKNOWN`. Combined with shape-⊤ (`null` dims), the full-⊤ marker `(DType.UNKNOWN, null)` would be pinned.

## Candidate APIs to Probe

- `tf.py_function(...)` with no `Tout` spec.
- `tf.cast(t, dtype=<dynamic>)` where the target dtype is opaque.
- A parameter reached only via `**kwargs` from an unanalyzable call site.
- Composition with `tf.numpy_function` or similar Python-bridge APIs.

Each candidate needs Ariadne probing to confirm it produces dtype-⊤ rather than a hardcoded default.

## Scope

Read-only investigation first (probe each candidate, log what `Parameter.getTensorTypes()` returns), then add the fixture and assertion. If no candidate reliably produces dtype-⊤, file an Ariadne tracker for an API that should but doesn't (or accept that the dtype-⊤ axis is rarely exercised in practice).

## Cross-Refs

- #486 (lattice violation, original).
- `com.ibm.wala.cast.python.ml/CONTRIBUTING.md` § `Dtypes—getDefaultDTypes` for the contract.

PR #492, opened 2026-05-16
Document lattice ⊤/⊥ contract in `Parameter.inferTensorTypes`; pin behavior

### LINKED ISSUE #486: Lattice violation
We're missing a check for `null` here. According to https://github.com/ponder-lab/ML/blob/master/CONTRIBUTING.md#shapesgetdefaultshapes, `null`:

> ⊤—the generator produces a tensor, but its shape cannot be determined.

https://github.com/ponder-lab/Hybridize-Functions-Refactoring/blob/061dc111c1df2277c3fd8adc4db53e8d3049fd2b/edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java#L582

I even see the following comment that states that we need to honor the lattice contract:

https://github.com/ponder-lab/Hybridize-Functions-Refactoring/blob/061dc111c1df2277c3fd8adc4db53e8d3049fd2b/edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Parameter.java#L452-L480

This issue is to change our code to honor this contract.

khatchad:
## Summary

Closes #486. Supersedes #490 (Copilot bot PR; see [#490 comment-4466849703](https://github.com/ponder-lab/Hybridize-Functions-Refactoring/pull/490#issuecomment-4466849703) for verdict).

## What the Lattice Violation Actually Was

`Parameter.inferTensorTypes`'s existing implementation already preserves the wala/ML lattice contract correctly:

- ⊥ ("not a tensor"): no matching `TensorVariable`, no entry from the iterator → empty `Set<TensorType>`.
- ⊤ ("tensor with unknown shape"): Ariadne emits `TensorType` instances with `null` dims directly, per the `com.ibm.wala.cast.python.ml/CONTRIBUTING.md` `getTensorTypes` section ("Shapes and dtypes are orthogonal. When the shape is unknown but the dtype is known, `getTensorTypes` emits `TensorType` instances with `null` dims so dtype information is preserved.").

The violation was on the **documentation** side—the Javadoc claimed ⊤ and ⊥ were indistinguishable from the iterator's output, which is false: they're trivially distinguished by `isEmpty()` (⊥) vs. `anyMatch(tt -> tt.getDims() == null)` (⊤).

## Changes

- **`Parameter.inferTensorTypes` Javadoc**: rewrite to describe the actual contract (orthogonal shape/dtype, ⊤ marker is `getDims() == null`, callers that need to distinguish ⊤ from concrete must test for it).
- **Keep `IllegalStateException`** for the `tensorVariable == null` branch: `TensorTypeAnalysis.iterator` filters non-null, non-empty state, so this branch is unreachable under the current Ariadne contract; the throw surfaces a future invariant violation rather than silently coercing. (#490 had silenced this without a reproducer.)
- **`hybridize.target` Ariadne pin** bumped 0.43.0 → 0.44.0. No behavior change required for this commit; aligns with the in-flight ecosystem.

## Regression Tests

- `testInferredTensorTypesUnknownShapeTop`: `tf.keras.Input(shape=json.loads(...))` produces a tensor whose shape Ariadne cannot trace through `json.loads`. Asserts a shape-⊤ marker (`getDims() == null`) is present; dtype-agnostic (Ariadne supplies the `tf.keras.Input` default `float32`, consistent with the spec's orthogonality guarantee).
- `testInferredTensorTypesBottomNotTensor`: `f(5)` with a non-tensor argument. Asserts the empty set (⊥).

## Out of Scope

A full-⊤ fixture (both dtype and shape unknown, i.e., `TensorType(DType.UNKNOWN, null)`) is filed as a follow-up at #491. The candidate APIs that might produce dtype-⊤ (`tf.py_function` with no `Tout`, `tf.cast` with dynamic dtype, etc.) need independent Ariadne probing; this PR pins the shape-⊤ axis only.

## Verification

`./mvnw verify` green locally—473 tests, 0 failures, 0 errors, 11 skipped (baseline 471, +2 for the new fixtures).


edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -8074,4 +8074,57 @@ public void testInferredTensorTypes() throws Exception {
...
+		List<Parameter> parameters = function.getParameters();
+		assertNotNull(parameters);
+		assertEquals(1, parameters.size());
+
+		Parameter t = parameters.get(0);
+		assertEquals("t", t.getName());
+
+		Set<TensorType> inferred = t.getTensorTypes();
+		assertNotNull(inferred);
+		assertFalse("Lattice ⊤ must not collapse to ⊥.", inferred.isEmpty());
```
khatchad: `TensorType` consists of both shape and dtype. The [lattice documentation](https://github.com/ponder-lab/ML/blob/ebcfce5b79608295762e9e50e6c6f2adfca7c743/CONTRIBUTING.md#tensor-type-generators) says that empty set is for shape and dtype *individually* representing "provably not a tensor." How does checking for empty set on the entire `TensorType` correspond to that?

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -8074,4 +8074,57 @@ public void testInferredTensorTypes() throws Exception {
...
+		List<Parameter> parameters = function.getParameters();
+		assertNotNull(parameters);
+		assertEquals(1, parameters.size());
+
+		Parameter x = parameters.get(0);
+		assertEquals("x", x.getName());
+
+		Set<TensorType> inferred = x.getTensorTypes();
+		assertNotNull(inferred);
+		assertTrue("Lattice ⊥ must be the empty set.", inferred.isEmpty());
```
khatchad: Same deal here. It's not that the `TensorType` is empty it's that the constituent shape and dtype sets are empty according to https://github.com/ponder-lab/ML/blob/master/CONTRIBUTING.md#tensor-type-generators. Thoughts?

hybridize.target
```diff
@@ -57,7 +57,7 @@
 				<dependency>
 					<groupId>com.ibm.wala</groupId>
 					<artifactId>com.ibm.wala.cast.python.ml</artifactId>
-					<version>0.43.0</version>
+					<version>0.44.0</version>
```
khatchad, line 60: This should probably be in a separate PR that we merge first.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
khatchad: Agreed—my framing collapsed two distinct levels. Rewrote `Parameter.inferTensorTypes` Javadoc to be explicit: the lattice contract lives at the generator level (`getDefaultShapes`/`getDefaultDTypes`), not at the `TensorTypeAnalysis` iterator level. The TensorType-level markers that *do* propagate through the iterator are `getDims() == null` (shape-⊤) and `getDType() == DType.UNKNOWN` (dtype-⊤), orthogonal per the `Tensor Types—getTensorTypes` note. The test now asserts shape-⊤ marker observability (`getDims() == null`) at the TensorType level, and its Javadoc says so explicitly. Pushed in d0dabb3.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
khatchad: Same framing fix. The empty `Set<TensorType>` Hybridize observes here is the iterator no-entry signal, which collapses three generator-side conditions—no bound `TensorVariable`, variable-level ⊤ (`state == null`), and variable-level ⊥ (`state.isEmpty()`)—into one indistinguishable observable. The test name keeps the "Bottom" framing (the *scenario* is a non-tensor argument), but the Javadoc now spells out that the empty-set observation is not directly equivalent to the generator-side lattice ⊥, and that finer distinctions would require a richer Ariadne-side query. Pushed in d0dabb3.

hybridize.target
khatchad: Done. Ariadne 0.44.0 bump reverted from this PR in d0dabb3; refiled as #493. Once #493 merges, this PR rebases.

PR #496, opened 2026-05-16
Hoist `inferTensorTypes` to fire unconditionally on `isTensorTyped` entry

khatchad:
## Summary

Closes #495 (cache hole).

After #488/#489 moved the type-hint/Ariadne/container classification into `Parameter.isTensorTyped`, the `inferTensorTypes` call that populates the per-Parameter tensor-types cache lived only inside Phase 2 (Ariadne query). Three paths bypassed it:

- Self parameters (early return).
- Type-hint-classified parameters with `followTypeHints` enabled (Phase 1 returns true before Phase 2).
- Functions with empty call-graph nodes (Phase 2 already gated on `!nodes.isEmpty()`).

Consumers reading `Parameter.getTensorTypes()` (no-arg, introduced #480) silently saw an empty set in these cases—dropping type-hint-classified tensor parameters from `Function.inferInputSignature`'s output even when Ariadne had a concrete `TensorType` from the call site. The self and empty-nodes cases are benign (excluded explicitly by `inferInputSignature`); the type-hint case is the load-bearing leak.

## Change

Move the `inferTensorTypes(tensorAnalysis)` call from inside Phase 2 to right after the `isSelf` check, still gated on `!nodes.isEmpty()`. Phase 1 (type hints) and Phase 2 (Ariadne cache read) now both observe a consistent cache state. Phase 2 reads `getTensorTypes()` instead of re-invoking `inferTensorTypes`.

Self parameters remain excluded from the cache (no semantic need; `inferInputSignature` filters them via `isSelf()`).

## Regression Test

`testInferredTensorTypesUnderFollowTypeHints`: `experimental_follow_type_hints=True` on a function with a `tf.Tensor` type hint, called with `tf.constant([1.0, 2.0])`. Asserts the cache contains Ariadne's call-site `TensorType(FLOAT32, [NumericDim(2)])`.

Verified failing on `main` before the hoist:

```
java.lang.AssertionError: Cache must be populated from Ariadne's call-site classification under followTypeHints.
expected:<[{[D:Constant,2] of float32}]> but was:<null>
```

Passes after the hoist.

## Verification

`./mvnw verify` green locally: 472 tests, 0 failures, 0 errors, 11 skipped (+1 from baseline for the new regression test).

### DIFF
--- a/edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
+++ b/edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
@@ -50,6 +50,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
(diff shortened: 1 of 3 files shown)


edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -8074,4 +8075,36 @@ public void testInferredTensorTypes() throws Exception {
...
 	}
+
+	/**
+	 * Regression test for #495: under `experimental_follow_type_hints=True` with a `tf.Tensor` type hint, the per-Parameter tensor-types
+	 * cache must be populated even though `Parameter.isTensorTyped`'s Phase 1 (type hints) returns true before Phase 2 (Ariadne query)
+	 * runs. Without the hoist landed in #496, `param.getTensorTypes()` returned the empty default for type-hint-classified parameters even
+	 * when Ariadne had a concrete `TensorType` from the call site.
+	 */
+	@Test
+	public void testInferredTensorTypesUnderFollowTypeHints() throws Exception {
```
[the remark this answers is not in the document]
khatchad: Already addressed—the PR description was updated when the regression test landed (commit 9d50b73). Current body lists `testInferredTensorTypesUnderFollowTypeHints` in the "Regression Test" section and reports 472 tests (+1 from baseline). Copilot reviewed an older revision.

edu.cuny.hunter.hybridize.tests/test cases/edu/cuny/hunter/hybridize/tests/HybridizeFunctionRefactoringTest.java
```diff
@@ -8074,4 +8075,36 @@ public void testInferredTensorTypes() throws Exception {
...
+		List<Parameter> parameters = function.getParameters();
+		assertNotNull(parameters);
+		assertEquals(1, parameters.size());
+		Parameter t = parameters.get(0);
+		assertEquals("t", t.getName());
+
+		// The load-bearing assertion: the cache is populated from the call site's `tf.constant([1.0, 2.0])`, even though Phase 1's
+		// type-hint hit causes `isTensorTyped` to return true before Phase 2 reads the cache. Without the hoist, this assertion fails
+		// (the cache stays at the empty default).
+		TensorType expected = new TensorType(DType.FLOAT32.name().toLowerCase(Locale.ROOT), List.of(new TensorType.NumericDim(2)));
```
khatchad: Why not use the new `TensorType` ctor that takes a `DType`?

khatchad: Done in 321c6d5. Switched to `new TensorType(DType.FLOAT32, List.of(new NumericDim(2)))` — dropped the `name().toLowerCase(Locale.ROOT)` workaround that was needed before Ariadne 0.44.0. Also imported `TensorType.NumericDim` directly so the `TensorType.` qualifier is gone on both call sites in this file, and dropped the now-unused `java.util.Locale` import.

Issue #111, opened 2022-12-05
Look up values of hybridization parameters


khatchad:
We just have existence right now. Related to #30.

If the argument isn't present, we must use the default value. But, how will we know which signature to use (which version?)?

PR #514, opened 2026-05-19
Replace JavaBean-style `is*ParamExists` accessors with `has*Param`

khatchad:
## Summary

- Renames eight `*ParamExists` boolean fields + `is*ParamExists()` accessors on `Function.HybridizationParameters`. Following the `tensor`/`isTensor()` pattern established in #513: field is noun-form, accessor is predicate-form, the two names are distinct so no shadowing
- Drops the unjustified `is*` JavaBean prefix from accessors—`Function` is not exposed via JavaBean introspection, DI, or property serialization, so the convention provides no value and the doubled-predicate reading is the only side effect
- Fixes a pre-existing typo (`experimenta` → `experimental`) and normalizes the abbreviated `OptParam` accessor to `OptionsParam` (matching the field's full word)
- `getHasTensorParameter()` is intentionally untouched; it returns `Boolean` (tri-state), and the JavaBean `is*` rule applies only to primitive `boolean`. Filing a separate cleanup if desired

Closes #512

## Renames

| Field (was) | Field (now) | Accessor (was) | Accessor (now) |
|---|---|---|---|
| `autoGraphParamExists` | `autoGraphParam` | `isAutoGraphParamExists()` | `hasAutoGraphParam()` |
| `experimentaFollowTypeHintsParamExists` | `experimentalFollowTypeHintsParam` | `isExperimentalFollowTypeHintsParamExists()` | `hasExperimentalFollowTypeHintsParam()` |
| `experimentalAutographOptionsParamExists` | `experimentalAutographOptionsParam` | `isExperimentalAutographOptParamExists()` | `hasExperimentalAutographOptionsParam()` |
| `experimentalImplementsParamExists` | `experimentalImplementsParam` | `isExperimentalImplementsParamExists()` | `hasExperimentalImplementsParam()` |
| `funcParamExists` | `funcParam` | `isFuncParamExists()` | `hasFuncParam()` |
| `inputSignatureParamExists` | `inputSignatureParam` | `isInputSignatureParamExists()` | `hasInputSignatureParam()` |
| `jitCompileParamExists` | `jitCompileParam` | `isJitCompileParamExists()` | `hasJitCompileParam()` |
| `reduceRetracingParamExists` | `reduceRetracingParam` | `isReduceRetracingParamExists()` | `hasReduceRetracingParam()` |

## Test Plan

- [x] `./mvnw verify`: 492 tests, 0 failures, 0 errors, 11 skipped
- [x] Spotless passes (pre-commit hook)

### DIFF
--- a/edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
+++ b/edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
@@ -172,42 +172,42 @@ public class Function {
 		/**
 		 * True iff this {@link Function}'s {@link decoratorsType} has parameter autograph.
 		 */
-		private boolean autoGraphParamExists;
+		private boolean hasAutoGraphParam;
 
 		/**
 		 * True iff this {@link Function}'s {@link decoratorsType} has parameter experimental_follow_type_hints.
 		 */
-		private boolean experimentaFollowTypeHintsParamExists;
+		private boolean hasExperimentalFollowTypeHintsParam;
 
 		/**
 		 * True iff this {@link Function}'s {@link decoratorsType} has parameter experimental_autograph_options.
 		 */
-		private boolean experimentalAutographOptionsParamExists;
+		private boolean hasExperimentalAutographOptionsParam;
 
 		/**
 		 * True iff this {@link Function}'s {@link decoratorsType} has parameter experimental_implements.
 		 */
-		private boolean experimentalImplementsParamExists;
+		private boolean hasExperimentalImplementsParam;
 
 		/**
 		 * True iff this {@link Function}'s {@link decoratorsType} has parameter func.
 		 */
-		private boolean funcParamExists;
+		private boolean hasFuncParam;
 
 		/**
 		 * True iff this {@link Function}'s {@link decoratorsType} has parameter input_signature.
 		 */
-		private boolean inputSignatureParamExists;
+		private boolean hasInputSignatureParam;
 
 		/**
 		 * True iff this {@link Function}'s {@link decoratorsType} has parameter jit_compile.
 		 */
-		private boolean jitCompileParamExists;
+		private boolean hasJitCompileParam;
 
 		/**
 		 * True iff this {@link Function}'s {@link decoratorsType} has parameter reduce_retracing.
 		 */
-		private boolean reduceRetracingParamExists;
+		private boolean hasReduceRetracingParam;
 
 		private void computeParameterExistance() {
 			// Use the hybrid decorator cached by `computeHybridization` (#118). That method already iterated every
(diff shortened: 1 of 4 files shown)


edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -172,42 +172,42 @@ public class HybridizationParameters {
 		/**
 		 * True iff this {@link Function}'s {@link decoratorsType} has parameter autograph.
 		 */
-		private boolean autoGraphParamExists;
+		private boolean hasAutoGraphParam;
```
khatchad: Why not just `autoGraphParam` instead? Isn't the accessor usually named `hasAutoGraphParam()`?

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -247,99 +247,99 @@ private void computeParameterExistance() {
...
 		}
 
 		/**
 		 * True iff this {@link Function}'s {@link decoratorsType} has parameter autograph.
 		 *
 		 * @return True iff this {@link decoratorsType} has parameter autograph.
 		 */
-		public boolean isAutoGraphParamExists() {
-			return this.autoGraphParamExists;
+		public boolean hasAutoGraphParam() {
```
khatchad, line 278: It's weird to have an accessor method with the same name as the field.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
khatchad: > Why not just `autoGraphParam` instead? Isn't the accessor usually named `hasAutoGraphParam()`?

Same shape as #513's `tensor`/`isTensor()`: noun-form field, predicate-form accessor, distinct names. Applied in 04142cd7 across all eight fields.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
khatchad: > It's weird to have an accessor method with the same name as the field.

Same field/method shadowing pattern removed from `Parameter` in #511/#513. Fields renamed to noun form in 04142cd7; accessors retain `has*Param()`.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -172,42 +172,42 @@ public class HybridizationParameters {
 		/**
 		 * True iff this {@link Function}'s {@link decoratorsType} has parameter autograph.
 		 */
-		private boolean autoGraphParamExists;
+		private boolean autoGraphParam;
```
khatchad, line 175: I believe that the reason we used "`*Exists`" is because we were planning to try to resolve the actual values. But, we can always change these again when that time comes. In fact, they will probably disappear at that time.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
```diff
@@ -247,99 +247,99 @@ private void computeParameterExistance() {
...
 
 		/**
-		 * Set the appropriate {@code *ParamExists} field for the given {@code tf.function} parameter name. Recognizes both current names
-		 * and the deprecated aliases ({@code experimental_compile} → {@code jit_compile}, {@code experimental_relax_shapes} →
+		 * Set the appropriate {@code *Param} field for the given {@code tf.function} parameter name. Recognizes both current names and the
+		 * deprecated aliases ({@code experimental_compile} → {@code jit_compile}, {@code experimental_relax_shapes} →
 		 * {@code reduce_retracing}). Unknown names are silently ignored; they may belong to a future TF version we don't model yet.
 		 */
 		private void markParamExists(String paramName) {
 			if (paramName.equals(FUNC))
```
khatchad, line 255: Probably can use `switch` here.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
khatchad: > I believe that the reason we used "`*Exists`" is because we were planning to try to resolve the actual values. But, we can always change these again when that time comes. In fact, they will probably disappear at that time.

Helper-name vocabulary cleanup applied independently of #111 at #517.

edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
khatchad: > Probably can use `switch` here.

The ten parameter-name constants are all `private static final String` (lines 140–158), so an arrow-form switch maps cleanly:

```java
switch (paramName) {
    case FUNC -> this.funcParam = true;
    case INPUT_SIGNATURE -> this.inputSignatureParam = true;
    case AUTOGRAPH -> this.autoGraphParam = true;
    case JIT_COMPILE, EXPERIMENTAL_COMPILE -> this.jitCompileParam = true;
    case REDUCE_RETRACING, EXPERIMENTAL_RELAX_SHAPES -> this.reduceRetracingParam = true;
    case EXPERIMENTAL_IMPLEMENTS -> this.experimentalImplementsParam = true;
    case EXPERIMENTAL_AUTOGRAPH_OPTIONS -> this.experimentalAutographOptionsParam = true;
    case EXPERIMENTAL_FOLLOW_TYPE_HINTS -> this.experimentalFollowTypeHintsParam = true;
    default -> { } // unknown names silently ignored (per existing Javadoc)
}
```

Tracked at #516 as out-of-scope for #514.

PR #608, opened 2026-06-11
Overwrite a disagreeing `input_signature` (RECONFIGURE modify path)

khatchad:
## Summary

Completes the `RECONFIGURE` transformation: when a hybrid `@tf.function` already supplies an `input_signature`, compare it against the inferred one and respond per the relation. The add-when-absent half landed earlier (#563); this is the modify half (#596).

The comparison is a per-parameter, per-axis partial order on the tensor-type lattice: a wildcard (`None`/dynamic/symbolic dimension, unknown-rank shape, or `UNKNOWN` dtype) is the most general value, a concrete value is more specific, and two distinct concrete values (or shapes of different rank) are incomparable.

| Relation | Action | Status |
|---|---|---|
| supplied strictly tighter than inferred | overwrite with inferred | informational |
| supplied strictly broader | preserve | informational |
| incomparable | overwrite with inferred | warning |
| agreement | no-op | — |

## Changes

- `PreconditionSuccess.P5` for the overwrite path. `check()` selects `RECONFIGURE`+`P5` for the overwrite relations, preserves for the broader relation, and no-ops on agreement — gated on the `inferInputSignatures` flag (default off), side-effect-free, non-recursive, and an emittable inferred signature, so the default precondition matrix is unchanged.
- `reconfigure()` extended from insertion to in-place replacement of the existing `input_signature` value (its AST node is retained during parameter parsing so its source span can be located).
- `InputSignature.Relation` + `relate()` implement the lattice comparison.

## Tests

`InputSignature.relate()` is unit-tested directly (deterministic, no analysis) across all relations and edge cases (rank/dtype/parameter-count mismatch, wildcard equivalence). Integration fixtures cover one example per relation — overwrite (tighter, informational), overwrite (incomparable, warning), preserve (broader), agreement (no-op) — plus a flag-off guard, asserting the emitted source and the status severity. The overwrite fixtures intentionally carry a signature the call sites violate, so they are analyzed statically rather than executed. Full suite: 531 pass, 11 known-skipped.

Closes #596

### DIFF
--- a/edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
+++ b/edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/Function.java
@@ -9,6 +9,7 @@ import static edu.cuny.hunter.hybridize.core.analysis.PreconditionSuccess.P1;
 import static edu.cuny.hunter.hybridize.core.analysis.PreconditionSuccess.P2;
 import static edu.cuny.hunter.hybridize.core.analysis.PreconditionSuccess.P3;
 import static edu.cuny.hunter.hybridize.core.analysis.PreconditionSuccess.P4;
+import static edu.cuny.hunter.hybridize.core.analysis.PreconditionSuccess.P5;
 import static edu.cuny.hunter.hybridize.core.analysis.Refactoring.CONVERT_EAGER_FUNCTION_TO_HYBRID;
 import static edu.cuny.hunter.hybridize.core.analysis.Refactoring.OPTIMIZE_HYBRID_FUNCTION;
 import static edu.cuny.hunter.hybridize.core.analysis.Transformation.CONVERT_TO_EAGER;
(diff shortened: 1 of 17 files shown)


khatchad (commented):
The comparator and precondition wiring look clean and faithful.

- `InputSignature.relate()` is a correct per-parameter, per-axis product order: `combine()` makes `AGREEMENT` the identity and `INCOMPARABLE` absorbing, and a mixed `SUPPLIED_TIGHTER` + `SUPPLIED_BROADER` correctly becomes `INCOMPARABLE` (neither side is uniformly at least as specific). Wildcard handling is consistent across all three axes (any non-`NumericDim` dimension, null-rank shape, and `UNKNOWN` dtype as the top).
- The `Relation` → transformation/severity wiring in `check()` lines up: `SUPPLIED_TIGHTER`/`INCOMPARABLE` → `RECONFIGURE`+`P5` (informational / warning), `SUPPLIED_BROADER` → preserve (informational), `AGREEMENT` → no-op. Gating on `inferInputSignatures` keeps the default precondition matrix unchanged.
- Direct unit tests on `relate()` plus one integration fixture per relation and a flag-off guard give good coverage.

One minor, non-blocking note: a parameter-count mismatch is folded into `INCOMPARABLE` (→ overwrite + warning). Reasonable, but a mismatched-arity supplied signature is arguably malformed rather than merely incomparable; a distinct status could be more actionable if it ever shows up.

PR #614, opened 2026-06-11
Emit inferred `input_signature` and supplied-vs-inferred relation in the evaluator

khatchad:
Adds three per-function CSV columns to the evaluator, together characterizing each function's input-signature inference outcome across the corpus:

- `inferred input_signature`—the signature the refactoring computed, as a `tf.TensorSpec(...)` list.
- `input_signature relation`—when both an explicit and an inferred signature exist, how they relate under the per-parameter, per-axis order (agreement, supplied tighter, supplied broader, incomparable).
- `input_signature absence reason`—when inference ran but was blocked, why (non-tensor parameter, no shape/dtype evidence, heterogeneous dtype, dtype-⊤).

For a function the refactoring ran inference on, exactly one of the inferred content and the absence reason is populated. All three read the memoized inference result via the side-effect-free getters (`getInferredInputSignature()`, `getInferredInputSignatureAbsenceReason()`), so they leave the existing `status` column untouched and are blank for functions where the refactoring never ran inference. The relation column reuses the lattice comparison from #608; the absence reason reuses the typed result from #483.

Stacked on #483 → #588 → #612 → #608. The `supplied input_signature` content column (#610, off `main`) is independent; when both land they touch adjacent entries in the column lists and merge trivially.

Closes #613

### DIFF
--- a/edu.cuny.hunter.hybridize.eval/src/edu/cuny/hunter/hybridize/eval/handlers/EvaluateHybridizeFunctionRefactoringHandler.java
+++ b/edu.cuny.hunter.hybridize.eval/src/edu/cuny/hunter/hybridize/eval/handlers/EvaluateHybridizeFunctionRefactoringHandler.java
@@ -515,11 +515,13 @@ public class EvaluateHybridizeFunctionRefactoringHandler extends EvaluateRefacto
 		printer.print(hybridizationParameters == null ? null : hybridizationParameters.hasJitCompileParam());
 		printer.print(hybridizationParameters == null ? null : hybridizationParameters.hasReduceRetracingParam());
 
-		// The inferred signature the refactoring computed for this function, how it relates to an explicitly supplied one, and—when no
-		// signature was inferred—why. All read the memoized inference result without recomputing, so they leave the status untouched. When
-		// the refactoring did not run inference on a function, all three are blank. When it did, exactly one of the inferred-content and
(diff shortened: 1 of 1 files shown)


edu.cuny.hunter.hybridize.eval/src/edu/cuny/hunter/hybridize/eval/handlers/EvaluateHybridizeFunctionRefactoringHandler.java
```diff
@@ -514,6 +515,17 @@ private static void printFunction(CSVPrinter printer, Function function) throws
 		printer.print(hybridizationParameters == null ? null : hybridizationParameters.hasJitCompileParam());
 		printer.print(hybridizationParameters == null ? null : hybridizationParameters.hasReduceRetracingParam());
 
+		// The inferred signature the refactoring computed for this function, how it relates to an explicitly supplied one, and—when no
+		// signature was inferred—why. All read the memoized inference result without recomputing, so they leave the status untouched. When
+		// the refactoring did not run inference on a function, all three are blank. When it did, exactly one of the inferred-content and
+		// absence-reason columns is populated: the content when a signature was inferred, the reason when inference was blocked. (The
+		// inferred-content column is thus blank both when inference never ran and when it ran but was blocked.)
```
khatchad: Let's use a multi-line comment and not break lines.

PR #618, opened 2026-06-17
Emit `RaggedTensorSpec` for ragged tensor parameters

khatchad:
Input-signature inference collapsed a parameter's ragged dimension (Ariadne's `RaggedDim`) to a symbolic wildcard and emitted a dense `tf.TensorSpec`, discarding the raggedness. Now:

- `inferSpec` preserves a consensus `RaggedDim` through its per-dim reduction instead of collapsing it.
- `InputSignature.toTensorSpecList` emits `tf.RaggedTensorSpec` for a parameter with a ragged dimension (the ragged positions still render as `None` in the shape).
- `InputSignature.requiredSpecTypeNames` reports the spec-type constructors a signature needs (`TensorSpec`/`RaggedTensorSpec`), and `computeInputSignatureKeyword` gates emission on them being reachable—so a ragged signature under a named import that lacks `RaggedTensorSpec` is skipped rather than emitting a `NameError`-raising decorator, matching the existing dtype-reachability guard (#585).

Confirmed empirically that Ariadne 0.47.0 types a `tf.RaggedTensor` parameter with a `RaggedDim` (new test `testInferRaggedTensorSpec`: `tf.ragged.constant([[1,2,3],[4,5]])` → `[tf.RaggedTensorSpec(shape=(2, None), dtype=tf.int32)]`).

The 29 precision-audit corpus pins and one formatter unit pin that asserted the old symbolic-wildcard collapse are updated to expect the preserved `RaggedDim`/`RaggedTensorSpec` (their param-layer already pinned `RaggedDim`; only the inferred-layer expectation moved). Full suite green (558).

Follow-up: auto-injecting `RaggedTensorSpec` into an import-less file (so ragged emits there too, rather than safely skipping)—analogous to the `TensorSpec` auto-inject (#574).

Stacked on #483 → #588 → #612 → #608.

Closes #524

### DIFF
--- a/edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/InputSignature.java
+++ b/edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/InputSignature.java
@@ -22,6 +22,12 @@ import com.ibm.wala.cast.python.ml.types.TensorType.RaggedDim;
  */
 public record InputSignature(List<TensorType> parameterTypes) {
 
+	/** The {@code tf.TensorSpec} constructor name, emitted for a dense tensor parameter. */
+	private static final String TENSOR_SPEC = "TensorSpec";
+
+	/** The {@code tf.RaggedTensorSpec} constructor name, emitted for a ragged tensor parameter (#524). */
+	private static final String RAGGED_TENSOR_SPEC = "RaggedTensorSpec";
+
 	/**
 	 * Format this signature as a Python source-code expression suitable for the {@code input_signature=} keyword argument of
 	 * {@code @tf.function(...)}. Each parameter's {@link TensorType} renders as {@code tfPrefix + "<SpecType>(shape=(...), dtype=" +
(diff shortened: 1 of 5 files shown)


edu.cuny.hunter.hybridize.core/src/edu/cuny/hunter/hybridize/core/analysis/InputSignature.java
```diff
@@ -61,12 +63,40 @@ public String toTensorSpecList(String tfPrefix) {
...
+	 * {@code tf.RaggedTensor}), otherwise {@code TensorSpec}. The bare name without any module prefix.
+	 *
+	 * @param t The parameter's {@link TensorType}.
+	 * @return {@code "RaggedTensorSpec"} or {@code "TensorSpec"}.
+	 * @see <a href="https://github.com/ponder-lab/Hybridize-Functions-Refactoring/issues/524">Issue 524</a>
+	 */
+	private static String specTypeName(TensorType t) {
+		List<Dimension<?>> dims = t.getDims();
+		boolean ragged = dims != null && dims.stream().anyMatch(RaggedDim.class::isInstance);
+		return ragged ? "RaggedTensorSpec" : "TensorSpec";
```
khatchad: Constants for `"RaggedTensorSpec"` and `"TensorSpec"`?

